PR #2, opened 2022-12-09
fix: parse whole line of comment

jgollenz:
The grammar was only capturing the hashtag that initiates the comment, but not the rest of the line.


fdncred:
Thanks! Hoping this lil repo will grow to some level of maturity so we can move it to the nushell repo.

PR #3, opened 2022-12-09
String Literals

jgollenz:
@fdncred I'm making baby steps here. There is much more to string literals, so I'll leave it as a draft for now. Right now, the rule is called `string_literal`, but only considers double-quoted string literals. Do you think we should do something like 

```js
string_literal: $ => choice(
	double_quoted_string,
	single_quoted_string,
...
```
?

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -404,11 +404,9 @@ module.exports = grammar({
 
         escape_sequence: $ => token.immediate(
             seq('\\',
-                choice(
-                    /[^xu]/,
+                choice( // nushell.sh/book/working_with_strings.html#double-quoted-strings
+                    /[bfrnt"'\/\\]/,
                     /u[0-9a-fA-F]{4}/,
-                    /u{[0-9a-fA-F]+}/,
-                    /x[0-9a-fA-F]{2}/
                 )
             )
         ),
--- a/grammar.js
+++ b/grammar.js
@@ -102,10 +102,10 @@ module.exports = grammar({
             // $.block_comment
         ),
 
-        line_comment: $ => token(seq(
+        line_comment: $ => token(prec(-1, seq(
             '#',
             /.*/,
-        )),
+        ))),
 
         // How lines can end
         line_terminator: $ => choice(
@@ -345,12 +345,12 @@ module.exports = grammar({
         ),
 
         string_literal: $ => seq(
-            alias(/b?"/, '"'),
+            token('"'),
             repeat(choice(
                 $.escape_sequence,
-                // $._string_content
+                /[^\\]/,
             )),
-            token.immediate('"')
+            token('"')
         ),
 
         integer_literal: $ => token(seq(
--- a/grammar.js
+++ b/grammar.js
@@ -344,13 +344,35 @@ module.exports = grammar({
             // $.float_literal,
         ),
 
-        string_literal: $ => seq(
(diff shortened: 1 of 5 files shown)


grammar.js
```diff
@@ -102,10 +102,10 @@ module.exports = grammar({
             // $.block_comment
         ),
 
-        line_comment: $ => token(seq(
+        line_comment: $ => token(prec(-1, seq(
```
jgollenz, line 105: I added this, because otherwise a literal like `"#hash"` will be parsed as a line comment.

grammar.js
```diff
@@ -345,12 +345,12 @@ module.exports = grammar({
         ),
 
         string_literal: $ => seq(
-            alias(/b?"/, '"'),
+            token('"'),
```
jgollenz, line 348: I might be mistaken, but running `let foo = b"test"` does not work with my version of nu. Am I overlooking something?

grammar.js
```diff
@@ -345,12 +345,12 @@ module.exports = grammar({
...
             repeat(choice(
                 $.escape_sequence,
-                // $._string_content
+                /[^\\]/,
```
jgollenz, line 351: Extracting this into it's own rule does not seem feasible, as the regex for the string content depends on the type of string we are parsing. This one does not allow any unescaped backslashes while e.g. single-quote strings don't allow single-quotes.

grammar.js
```diff
@@ -345,12 +345,12 @@ module.exports = grammar({
...
             )),
-            token.immediate('"')
+            token('"')
```
jgollenz, line 353: Using `immediate` made strings like `let foo = "bar  "` illegal.

grammar.js
fdncred: You are right. I stole a bunch of the current stuff from the rust tree sitter. It's a great resource that we can copy from.

fdncred:
Seems reasonable. I'd just reiterate that learning from others work is helpful. This is why most of the grammar here comes from the rust tree sitter.

jgollenz:
> learning from others work is helpful

Absolutely. I'm jumping around between the grammars for other languages I know. Unfortunately Rust is not one of them (yet)

grammar.js
```diff
@@ -382,11 +404,9 @@ module.exports = grammar({
 
         escape_sequence: $ => token.immediate(
             seq('\\',
-                choice(
-                    /[^xu]/,
+                choice( // nushell.sh/book/working_with_strings.html#double-quoted-strings
+                    /["'\\\/bfrnt]/,
```
fdncred, line 408: There's more nushell escapes than \f \r \n \t and i'm not sure what \/b is

test/corpus/string_literal.txt
```diff
@@ -0,0 +1,97 @@
+==================
+string literal
+==================
```
fdncred, line 3: Definitely love these tests. We should have tests for everything for sure!

jgollenz:
Perhaps we should merge the current functionality under `simple string literals`. The other stuff seems a bit more complex and I don't want to bloat this PR.

fdncred:
> Perhaps we should merge ...

I'm fine with that. Mark it ready when you're ready.

fdncred:
Yup, not sure if this is all "right" but I'm good with chipping away at the block one thing at a time. Thanks!

PR #8, opened 2023-04-28
add *neovim* instructions to the `README`

amtoine:
i had some trouble initially to install some *neovim* stuff...

@zioroboco has written a complete set of instructions and tips in their [`nu-ls.nvim`](https://github.com/zioroboco/nu-ls.nvim) :ok_hand: 

there is a particularly interesting part for the `treesitter-nu` repo, i think, namely the [*Syntax highlighting?* chapter](https://github.com/zioroboco/nu-ls.nvim#syntax-highlighting) :smirk: 

## if that is ok with everyone, @1Kinoti and @zioroboco, i propose to move that section over to `nushell/treesitter-nu` :yum: 

this PR is the first part of this, i.e. migrating from @zioroboco's repo to here :relieved: 

> **Note**
> see the [new `README`](https://github.com/amtoine/tree-sitter-nu/tree/feature/add-neovim-instructions-to-the-readme#installation-guide)
> and i've added some more formatting, e.g. github notes and warnings, and a `nushell` snippet to install the `highlights.scm` file


amtoine:
> I'm kind of thinking that these editor-specific items should be in their own `.md` file. So, then we can have `spacevim`, `neovim`, `somebody_else_vim`, `helix`, etc. Then on the readme.md, just link to them. Thoughts?
>
we can definitely do that :ok_hand: 

i left that in the README because there is only one, but i can start in the right direction in this PR :relieved: 
let me do it :+1:

amtoine:
same as before, but with a separate file :ok_hand:

PR #11, opened 2023-05-05
fix `use` parsing precedence issue

1Kinoti:
this pr fixes a precedence issue when parsing a `use` declaration.

before
```nu
use "file.nu" some-cmd
------^------ ----^---
     use      pipeline
```

now
```nu
use "file.nu" some-cmd
----------^-----------
         use
```

it also allows an optional semicolon at the end of `use`, `def`, `def-env`, `export` and `alias`

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -43,12 +43,15 @@ module.exports = grammar({
 
         /// Top Level Items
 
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks @1Kinoti

PR #12, opened 2023-05-06
add query files from LhKipp/nvim-nu

amtoine:
i've added the queries from https://github.com/LhKipp/nvim-nu :+1: 

> **Note**
> we have a lot more queries here, so i did not get them :relieved: 

### Command used
```nu
[folds injections locals] | each {|file|
    http get $"https://raw.githubusercontent.com/LhKipp/nvim-nu/main/queries/nu/($file).scm"
    | save $"queries/($file).scm"
}
```

### DIFF
--- a/installation/neovim.md
+++ b/installation/neovim.md
@@ -47,7 +47,15 @@ let local = (
 let file = "highlights.scm"
 
 mkdir $local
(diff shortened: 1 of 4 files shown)


fdncred:
ya, we can't do this. LHKipp's is not MIT licensed. Sorry.

PR #13, opened 2023-05-07
update grammar

1Kinoti:
this is a very large pr that fixes a lot of issues with the parser. i know prs like these are discouraged. but since tree-sitter-nu is still in WIP status, i think it doesn't do much harm

# changes
1. tests - i added a sort of skeleton test suite that maybe expanded later

2. item termination - now items (statements and declarations) are terminated by newlines, semicolons or both. this makes it easier to deal with precedence

3. keywords with flags - keyword nodes like `do` and `for` that have flags are now parsed correctly

4. use of unquoted instead of identifier - previously arguments to `source` and `register` were parsed as identifiers and this was very limiting. so now we parse them as unquoted strings. this means we accumulate anything up to a whitespace

5. scope pattern - `use` and `hide` now had a left precedence which led to the scope pattern not being parsed correctly
```nu
use   file.nu     cmd
-^-   ---^---     -^-
 kw    module      ╰──  anything after the module is a scope pattern 
                        and was previously parsed as the start of a pipeline
```

6. where - the where used to only expect a binary expression but now the rhs is optional to cater for this case
```nu
ls | where $pred
```

7. cell-paths - now parenthesized expressions can take cellpaths
```nu
let span = (metadata $x).span
           -------^----  --^--
           parenthesized   ╰──── a cellpath
             expression
```
also cell-paths with numbers can now be parsed, solving the [**3rd**](https://github.com/nushell/tree-sitter-nu/raw/main/resources/treesitter-nu.readme.3.png) issue

8. inbuilt variables - `$in`, `$nu`, `$env` and `$nothing` are now parsed specially to enhance syntax highlighting

9. unquoted string regex - to improve parsing of unquoted strings, their regex pattern has been updated to separate them from flags and variables. this partially solves the [**4th**](https://github.com/nushell/tree-sitter-nu/blob/main/resources/treesitter-nu.readme.4.png) issue

10. `module` - the `module` keyword is now parsed along with its new `export`ed form
```nu
module foo {}
export module foo {}
```
11. syntax-highlighting - queries have been added for
  - binary operators of the `where` argument's binary expression
  - dates
  - the `:` key-value separator in records
  - enhanced parameter types and annotations

12. decls in closures - declarations (top level items are now parsed in closures)

13. short flags - short flags are now parsed as short flags as opposed to strings previously
```
some-cmd -f
          ╰──── was previously parsed as a string
``` 

... probably other minor changes


fdncred:
I think this is a great step forward. I especially like the corpus tests. When I originally started tree-sitter-nu I was always doing tests like this. It's really one of the few ways to make sure you don't break something else because at times the grammar can be quite delicate and fragile.

We do want to keep PRs smaller but while we're in the early stages, trying to figure everything out still, this is probably ok for now.  Good work!

fdncred:
Unrelated but at some point I'd like to see some general instructions on how to contribute and test. Not necessarily a complete contributing guide, but in my original readme, I laid out a few steps on where to make changes, how to validate those changes worked, etc.

fdncred:
Your ts.nu may have a bug too. I get this
```
use ts.nu *  
Error: nu::parser::variable_not_found

  × Variable not found.
    ╭─[/Users/fdncred/src/tree-sitter-nu/ts.nu:36:1]
 36 │     if $debug {
 37 │         tree-sitter parse $file -d
    ·                           ──┬──
    ·                             ╰── variable not found.
 38 │     } else if $stat {
    ╰────
```

1Kinoti:
> Unrelated but at some point I'd like to see some general instructions on how to contribute and test. Not necessarily a complete contributing guide, but in my original readme, I laid out a few steps on where to make changes, how to validate those changes worked, etc.

@fdncred, i am actually writing a `CONTRIBUTING.md` where i will give all these instructions.

fdncred:
Thanks. I plastered in my rough notes from the original readme into the new one. Feel free to modify that.

PR #17, opened 2023-05-08
Treat shebangs as regular comments in the highlights query

zioroboco:
👋 😊

Shebang lines (`#!/some/binary --etc`) are a distinct type according to the grammar, but currently aren't included in the highlights query.

This change treats them as regular comments for the purpose of highlighting.


fdncred:
Let's land this and circle-back later if we want to change the shebang coloring.

PR #20, opened 2023-07-07
add support for table annotations

1Kinoti:
response to [nushell#9613](https://github.com/nushell/nushell/pull/9613)


fdncred:
nice! thanks!

PR #21, opened 2023-07-11
improve cellpaths

1Kinoti:
now we can parse

```nushell
$env.PATH
$list.1
$x.FOO.5
```

forgot this during the initial impl, but now i have added tests


fdncred:
Thanks @1Kinoti

PR #22, opened 2023-07-16
Unifying rules for command parsing

SirZenith:
Currently, there are two different rules for command parsing. One for sub commands like `str join`, another for simple command like `echo`.

This make application side difficult  to tell the difference between `echo this` and `str join`, cause they are both being parsed as sub command.

And this two rules also make command with three or more name segments failed to parse.

 - This pull request simplify those two rule into one, which always parse and only parse the first command segment as command name, everything else that follows are arguments.
  Applications can look into arguments on their own, to recognize long command with more than one name segment.
- Also this pull request rewrite part of pipeline rule, to allow user add new line before and after `|`.
- Add an extra argument to `ts test` command in `ts.nu` to allow test case name filtering (also added completion for that arg).


fdncred:
@SirZenith Thanks for this PR!

@1Kinoti Would you mind reviewing this PR and letting us know what you think?

1Kinoti:
@SirZenith thanks for this.

a few nitpicks and we'll be good go

grammar.js
```diff
@@ -616,16 +614,12 @@ module.exports = grammar({
...
-                            $.pipeline,
-                            BRACK().close_paren
-                        ),
-                        $.expr_parenthesized
-                    )
+                    seq(
+                        // ensure the expression immediately follows the
+                        // opening paren
+                        token.immediate(BRACK().open_paren),
+                        inline_pipeline($, BRACK().close_paren),
```
1Kinoti, line 621: not quite related to your pr, but i just found a bug

```nushell
(let x = 4; x) 
```

is rare but valid nu, so we should probably fix that
```suggestion
                        $._top_level_block,
                        BRACK().close_paren
```

grammar.js
```diff
@@ -907,21 +900,10 @@ module.exports = grammar({
 
         /// Commands
 
-        command: $ => prec.right(10, choice(
-            $.cmd_head,
-            $.cmd_head_sub,
-        )),
-
-        cmd_head_sub: $ => prec.right(2, seq(
+        command: $ => seq(
```
1Kinoti, line 903: nice!

corpus/stmt/let.nu
```diff
@@ -24,27 +22,25 @@ let x = 42;
...
-      (pipe_element
-        (command
-          (cmd_head_sub
-            (cmd_identifier)
-            (cmd_identifier)))))))
+    (val_number))
+  (pipeline
+    (pipe_element
+      (command
+        (MISSING cmd_identifier)))
```
1Kinoti, line 42: oops! this test is legal nu

ts.nu
1Kinoti, line 1: nice to see you updated `ts.nu` to have completions, but in the future maybe open a separate pr for that? also, this is a personal preference, but the main file functions should appear first, and helpers at the bottom ie, `ts gen`, `ts hl` and `ts test` should appear first here since they are the focus of this module.

1Kinoti (changes requested):

grammar.js
```diff
@@ -427,7 +430,7 @@ module.exports = grammar({
         _assignment_pattern: $ => seq(
             field("name", $._variable_name),
             PUNC().eq,
-            field("value", $.pipeline),
+            field("value", $._expression),
```
1Kinoti, line 433: starting from [#9589](https://github.com/nushell/nushell/pull/9589) and [#9658](https://github.com/nushell/nushell/pull/9658) 

```nushell
let x = ls | get name
# and
mut x = ls | get name
```

are legal

```nushell
const x = some const | pipeline
```

is not ([yet?](https://github.com/nushell/nushell/pull/9499)) legal nushell, but i think the parser should be robust enough to handle it
```suggestion
            field("value", $._pipeline),
```

1Kinoti (changes requested):

grammar.js
```diff
@@ -549,20 +552,15 @@ module.exports = grammar({
...
-        pipeline: $ => prec(-69, seq(
-            $.pipe_element,
-            repeat(seq(
-                PUNC().pipe,
-                $.pipe_element
-            )),
-        )),
 
-        pipe_element: $ => prec(-1, choice(
+        pipeline: $ => inline_pipeline($, $._terminator),
```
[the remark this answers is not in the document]
SirZenith: Quite strange though, I found pipeline rule must have an explicit terminating character to make the rule match greedily.
Some thing like [this](https://github.com/tree-sitter/tree-sitter/issues/1139).
With out terminating character, tree-sitter would prefer make `(echo this)` into 2 pipelines wrapped in parenthesis.
`inline_pipe` is here to make pipeline works in `expr_parenthesis`.

grammar.js
SirZenith: No, these let, mut, const assignment with pipeline are not legal nu statement yet. At least nu 0.82.0 on my machine does not accept those as valid statements.

ts.nu
SirZenith: Roger that.

grammar.js
SirZenith: But, yeah, maybe we can make our parser support that.

1Kinoti: we generally use the latest nushell main to add changes. if a syntax change is pushed to main we implement it way before the release. that way when a release is out, the parser already supports any new syntax

this change is recent, but i saw it coming :) that's why the parser already handles it. table annotations are also recent and have been already [implemented](https://github.com/nushell/tree-sitter-nu/pull/20)

i think by being one step ahead of the release, we can reduce surprises to people who use the parser for syntax highlighting

grammar.js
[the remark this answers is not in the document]
1Kinoti: we should probably add a test for that to prevent another regression? small changes can really make whole alternate universes in ts

grammar.js
SirZenith: I run test with this change locally and encounter an error, maybe we should leave this for later?

grammar.js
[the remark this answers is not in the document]
SirZenith: Good idea. Actually I have add bunch of test cases on my working branch, though many of them still can't pass.

Would it be OK to make a PR with only test case update, and even with ones that failed in it?

grammar.js
SirZenith: Done, I've change `_expression` back to `pipeline`, tests also get reverted to previous version.

grammar.js
1Kinoti: yeah, i agree. this wasn't part of your pr

grammar.js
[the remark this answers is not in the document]
1Kinoti: yeah this is okay, we need more tests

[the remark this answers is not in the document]
SirZenith: How about I make a separate directory (other than `corpus` and `test/corpus`) for those new tests?
Avoiding being run by CI, but let every body see what we should do at the same time.

1Kinoti:
@fdncred this is good to go. thanks again @SirZenith

SirZenith:
😄

grammar.js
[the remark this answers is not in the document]
1Kinoti: i like this :handshake:

1Kinoti (approved):

grammar.js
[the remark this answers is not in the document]
fdncred: I like the idea of having all the tests, even the ones that fail, so we can see some things we need to work on. However, the ones that fail may need to be ignored somehow or commented out until we can get them working. I'm up for whatever y'all think is the best way to proceed.

fdncred:
> @fdncred this is good to go. thanks again @SirZenith

Thanks @1Kinoti and @SirZenith. We appreciate the support!!

PR #26, opened 2023-07-22
add extern optional block

1Kinoti:
response to [#8956](https://github.com/nushell/nushell/pull/8956)


fdncred:
should parser.c be in .gitignore or is it important to always have the latest in the repo?

1Kinoti:
mmmh, i wonder. but removing it would need users, i.e editors to build it for themselves. `nvim-treesitter` can do that, but i wonder if others (helix and lapce) can

fdncred:
if people need it complete when downloaded, then it's fine to leave it, but if they're always going to rebuild, then it's probably not necessary. i'm leaning toward leaving it for now.

1Kinoti:
fwiw, tree-sitter-rust keep it in [tree](https://github.com/tree-sitter/tree-sitter-rust/blob/master/src/parser.c)

fdncred:
ya, we should follow their lead. Thanks!

PR #27, opened 2023-08-17
Helix support

NotLebedev:
Add installation instruction for helix editor and mention it in `README.md`.

Add `injections.scm` query to support  `TODO` and `FIXME` highlights in comments (see [PR discussion](https://github.com/helix-editor/helix/pull/7873#discussion_r1287216390)):

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@
 
 ## installation guide
 - for *neovim* users, please have a look at [`installation/neovim.md`](installation/neovim.md)
+- for *helix* users, please have a look at [`installation/helix.md`](installation/helix.md)
 
 ## contributing
 see [instructions](./CONTRIBUTING.md)
--- a/installation/helix.md
+++ b/installation/helix.md
@@ -1,0 +1,47 @@
+Helix has built-in support for tree-sitter grammars and installations from
+most sources have grammars for common languages (including nu!) built in.
+
+### Using built-in helix grammars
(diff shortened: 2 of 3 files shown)


fdncred:
oh, cool! I've been wondering about how to do that. Thanks!

fdncred:
I'm kinda wondering if we should just PR our tree-sitter-nu to helix so it's the default? I wonder if it's ready?

fdncred:
I just used the instructions with the latest commit-hash and it worked fine with 1 exception. I had to delete the two left over queries that we don't have. The ones from LHKipps versions. Once I did that, and copied the grammars to the right place, things worked correctly.

NotLebedev:
>I had to delete the two left over queries that we don't have. The ones from LHKipps versions.

I tried to explain this part in the warning at the end of [installation/helix.md](https://github.com/nushell/tree-sitter-nu/blob/786689b0562b9799ce53e824cb45a1a2a04dc673/installation/helix.md?plain=1#L32C1-L47C9). I hoped it would be clear, but maybe I need to rewrite it to be more clear?

fdncred:
Ya, I read that but what didn't occur to me was to delete those 2 scm files. I figured they'd just work with the other ones, even though you said they were incompatible. This a me issue, not an issue with your PR. I would've needed something like, "Delete your existing SCM files and copy the new ones each time you do this.". :)

PR #28, opened 2023-08-24
Fixes to block handling

NotLebedev:
Fix various issues related to how blocks are handled:
1. Blocks could not be entirely on one line and not empty and `( )` blocks, string interpolations and closures could not be multiline and have more then one command. Issue was in how blocks were handled. In `$.block` individual pipelines were separated by newline or `;` and last one should have had a newline before closing bracket. On the other side in other cases `inline_pipeline` was used directly and pipeline could terminate on `)` or `}` but only one pipeline per block was allowed (and not assignments, `use` `overlay use` e.t.c. were allowed at all, despite being usable in nu). To fix this `_last` variants of all rules that could end in pipeline were introduced. This rules optionally require `$._terminator` and thus one line blocks are now possible. `_block_body` rule encapsulating this logic is use in all block-like rules. 
  
    Adding a lot of `_last` rules is not pretty but I could not come up with a working grammar without them:
    * If all newlines are are ignored grammar could not be parsed unambiguously at all. E.g.
        ```nu
        ls | print
        ls | print
        ```
        is indistinguishable from `ls | print print | ls`. First prints result of `ls` two times. Second prints `print` and then result of `ls`
    * Making `$._terminator` always optional in `$.pipeline` splits pipelines in parts and for some reason no amound of tinkering with `prec.right` fixes this. But I am surprised by this, but my due to limited knowledge of tree-sitter behavior details I was not able to rectify this and settled at the hack with `_last` variants of rules.
2. Mutable variable assignments are still broken but less broken then before. Before this change all pipeline on right hand side of  mutable var assignment was used as `rhs`. Now correctly only first value is used. Hovewer the rest of pipeline is parsed with error because pipelines could not begin with just `|` (and IMO they should not be able to do this in nu. I will probably make an issue soon). Now behavior is somewhat similar but still not 1:1 to nushell parser
3. Match hands can now have blocks as expressions after `=>`. E.g.
    ```nu
    match $x {
        _ => { print qwe }
    }
    ```
    is now parsed correctly. Also empty blocks are now correctly parsed as block, not empty record.


NotLebedev:
Turns out I made a mistake. Did not notice it at first but `()` and `{}` blocks handle newline differently. In `{}` new line means end of command so
```nu
do {
print qwe
print rty
}
```
Will print
```
qwe
rty
```

But
```nu
(
print qwe
print rty
)
```
Will print
```
qwe
print
rty
```

Which means, in `()` command stretches until `;` not newline. This is cool, but it means I have to fix this, which means another set of rules 😐

Maybe I need to parametrize rules, like its done for `inlin_pipeline`. Any tips?

NotLebedev:
Found right tool for the job. Turns out people do some heavy meta-programming in these grammars. One example is grammar for C. [Here](https://github.com/tree-sitter/tree-sitter-c/blob/05f62b9f84e91a5bfb3e768c0d1271f20ac8ee35/grammar.js#L115) a special function is used that [returns](https://github.com/tree-sitter/tree-sitter-c/blob/05f62b9f84e91a5bfb3e768c0d1271f20ac8ee35/grammar.js#L770) parts of rules structure that are substituted with some suffix and additional rules. Seems like the right tool for this job. Seems lik it should not be too hard to implement and maintain later

fdncred:
@NotLebedev Are we ready to go with this? Doesn't look like @1Kinoti has bandwidth to review. We need more help with tree-sitter so we appreciate your contribution.

NotLebedev:
I was able to fix all problems I mentioned above. From my side everything is done, so we may proceed. It's unfortunate no review is coming

fdncred:
@amtoine what's the thumbs down about? Are you disapproving of this PR or just sad because 1Kinoti doesn't have much time?

fdncred:
Thanks @NotLebedev. Sorry this took so long.

amtoine:
yes i am sad that the PR got closed but this was a miss click, sorry for the confusion :innocent:

1Kinoti:
@NotLebedev, i am sorry i wasn't available to review this pr

PR #38, opened 2023-12-21
Fix parsing of export const

### LINKED ISSUE #34: Exports fail to highlight after first occurence
Zed
SCR-20231129-pcpz

Helix
SCR-20231129-pcts

```nu
export const GAMMA = 0.5772156649015329
export const E = 2.718281828459045
export const PI = 3.141592653589793
export const TAU = 6.283185307179586
export const PHI = 1.618033988749895
```

https://github.com/nushell/nushell/blob/main/crates/nu-std/std/math.nu

kit494way:
Fix #34

### DIFF
--- a/corpus/stmt/const.nu
+++ b/corpus/stmt/const.nu
@@ -62,3 +62,24 @@ export const x = 42
     (pipeline
       (pipe_element
         (val_number)))))
+
+=====
+const-005-exported-multiple
+=====
+
+export const GAMMA = 0.5772156649015329
+export const E = 2.718281828459045
+
+----
+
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks for this! I'm wondering if it's possible to have another test that specifically tests the #34 scenario where multiple in a row get recognized properly?

kit494way:
Add a test case for #34

fdncred:
Thanks!

JosephTLyons:
Thank you so much @kit494way, we're going to bump our internal version in Zed to include this commit :)

PR #39, opened 2023-12-23
Fix parsing of pipe at the end of line

### LINKED ISSUE #31: syntax highlighting gets confused
Trying out Zed with nu highlighting I found this in default_env.nu. Notice that after the pipe, it stops highlighting.

image

Looking at the `Syntax Tree` windows in Zed, there are a few `ERROR`s.

image

It would be nice to get this fixed in order to have a more pleasant experience in Zed.

kit494way:
Fix #31


fdncred:
cool. thanks for this!

PR #40, opened 2023-12-23
Fix parsing of unquoted 1-character string

### LINKED ISSUE #15: inconsistent behavior with unquoted strings
# Description

```nu
# this should parse as a pipeline with two elements
'nushell' | str contains n

# but it is parsed as two pipelines, with one ending at `contains`
'nushell' | str contains
n

# quoting the string solves it though
'nushell' | str contains 'n' # <- okay
```

kit494way:
Fix #15


fdncred:
you're on fire! thanks for all your help.

PR #43, opened 2023-12-24
Fix prettier configuration file name

kit494way:
Change the file name from .pretterrc to .prettierrc.
This pr does not change prettier settings, because .prettierrc is empty.
Before this change, if there is a .prettierrc in a parent directory, prettier read settings from the file and it may change the behavior of prettier.


amtoine (approved):
sounds legit :pray:

PR #48, opened 2023-12-27
Fix parsing of command

### LINKED ISSUE #45: Highlighting breaks on `alias` before `let`
The following breaks:

```nushell
alias less = bat
let foo = $'($env.HOME)'
```

It's fine when `alias` is removed.

### LINKED ISSUE #47: Fails to parse a multi-line command enclosed in parentheses
スクリーンショット 2023-12-26 203324

kit494way:
Fix #45 , Fix #47 

Under the current rule for parsing of command, whether the command is a subexpression or not, line breaks are allowed between command arguments.
This causes the parser to fail to correctly detect the end of a command or pipeline.
This pr changes the rules to allow line breaks between command arguments only when the command is a subexpression.

After this change, the highlights looks like follows
スクリーンショット 2023-12-27 215258
スクリーンショット 2023-12-27 215503

### DIFF
--- a/corpus/decl/alias.nu
+++ b/corpus/decl/alias.nu
@@ -45,3 +45,25 @@ export alias change-directory = cd;
       (pipe_element
         (command
           (cmd_identifier))))))
+
+=====
+alias-000-terminated-by-newline
+=====
+
+alias less = bat
+let foo = "foo"
+
+-----
+
+(nu_script
+  (decl_alias
+    (cmd_identifier)
+    (pipeline
+      (pipe_element
+        (command
+          (cmd_identifier)))))
+  (stmt_let
+    (identifier)
+    (pipeline
+      (pipe_element
+        (val_string)))))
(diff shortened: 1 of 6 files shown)


fdncred:
nice work! we appreciate all your efforts here.

amtoine:
nice :ok_hand: 


after updating, it looks much better :pray: 
i just have the whole content of the string interpolation that is white, i.e. `($env.HOME)`

fdncred:
@amtoine isn't that controlled by highlights.scm? I'm not positive but I thought it was.

amtoine:
> @amtoine isn't that controlled by highlights.scm? I'm not positive but I thought it was.
>
it might not be related to that PR yeah, because it's there in other places, i.e. `$env.HOME` never appear to highlight properly in my Neovim.

i did update both the parser and the query file :wink:

fdncred:
what i'm suggesting is that the highlights.scm file is very old and it probably needs a good updating too.

PR #55, opened 2024-01-02
Fix parsing of binary expression

### LINKED ISSUE #51: Parsing fails when the right-hand side of a binary expression is an unquoted string
スクリーンショット 2023-12-30 141552

kit494way:
Fix #51


fdncred:
Thanks!

PR #56, opened 2024-01-03
Fix parsing long flag

### LINKED ISSUE #54: Incorrect parsing for long param flag with dashes
First example:
```nu
#!/usr/bin/env nu

def main [
  --long-param-with-dashes
  --long_param_with_underscores
] {
  if $long_param_with_dashes {
    print "dash param"
  } else {
    print "no dash param"
  }

  if $long_param_with_underscores {
    print "underscore param"
  } else {
    print "no underscore param"
  }
}
```
Syntax highlighting is incorrect and there are multiple "param_short_flag" for the first param flag (with dashes):
tree-sitter-nu1

Second example (probably same cause but different effect):
```nu
#!/usr/bin/env nu

def main [
  --dry-run (-n)
] {
  if $dry_run {
    print "Dry run"
  } else {
    print "Real execution"
  }
}
```
There is now an error in the tree:
tree-sitter-nu2

kit494way:
Fix #54

### DIFF
--- a/corpus/decl/def.nu
+++ b/corpus/decl/def.nu
@@ -306,3 +306,47 @@ def test [
         (param_rest
           (identifier))))
     (block)))
+
+======
+def-017-long-flag-with-short-without-space
+======
+
+def test [
+  --long-flag(-l)
+] {}
+
+-----
+
+(nu_script
+  (decl_def
+    (cmd_identifier)
+    (parameter_bracks
+      (parameter
+        (param_long_flag
+          (identifier))
+        (flag_capsule
+          (param_short_flag))))
+    (block)))
+
+======
+def-018-long-flag-with-short-and-type
+======
+
+def test [
+  --long-flag(-l): int
+] {}
+
+-----
+
+(nu_script
+  (decl_def
+    (cmd_identifier)
+    (parameter_bracks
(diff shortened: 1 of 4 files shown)


corpus/decl/def.nu
```diff
@@ -110,3 +110,199 @@ def-env test []: nothing -> record<a: string, b: int> {}
...
+        (param_long_flag
+          (identifier))))
+    (block)))
+
+======
+def-008-long-flag-with-short
+======
+
+def test [
+  --long-flag (-l)
```
fdncred, line 138: does this require a space or will it parse like `--long-flag(-l)`?

fdncred:
Looks good but what about this type of flags?

```nushell
def test [
  --long-flag(-l): int
] {}
```

corpus/decl/def.nu
kit494way: A space is not required.

kit494way:
I will add more tests.

kit494way:
I added some tests.
I also fixed a same bug in a rule for long flags in command arguments.

fdncred:
thanks! i'm really enjoying all these fixes/additions. keep up the great work!

PR #57, opened 2024-01-03
Introduction of a simple plugin for nvim

aserowy:
Hi,

for easier integration into nvim, i added a plugin (tested with [folke/lazy.nvim](https://github.com/folke/lazy.nvim)) to the repository. To enable automatic resolution of queries, i had to move them to `queries/nu`.

The plugin sets the comment string, adds filetype detection and adds the parser to treesitters parser config. If you do not want to add a program specific plugin to this repository, please let me know. Thus, i ll host a seperated one for lazy devs like me :D

Kind regards
Alexander


kubouch:
Part of me is saying that it would be better to have it in a separate repository. But since tree sitter is intended primarily for editors, maybe it makes sense to have editor integrations here as well?

aserowy:
Long term should be a direct integration into the treesitter repo imho.

I created the pr to enable easy nvim support short term.

If your goal is to not integrate I would go the seperation of concerns route and create a repo for each editor. Thus, no plugin integration conflicts with others.

fdncred:
Could you also update [this markdown](https://github.com/nushell/tree-sitter-nu/blob/main/installation/neovim.md) with an example based on lazy.nvim?

fdncred:
thanks for updating the doc too. i think this makes things better.

fdncred:
FYI - This PR broke `tree-sitter highlight`. I'm hoping this fixes it https://github.com/nushell/tree-sitter-nu/pull/75

PR #61, opened 2024-01-04
Fix parsing of if/else and try/catch in parens

### LINKED ISSUE #58: A newline between `}` and `else` cause parse error
A newline between `}` and `else` is allowed only if it is parenthesized.
This is originally reported in this [comment](https://github.com/nushell/tree-sitter-nu/issues/52#issuecomment-1874017078).

スクリーンショット 2024-01-03 153748

kit494way:
Fix #58


fdncred:
Thanks!

PR #64, opened 2024-01-05
Change to parse composite types

kit494way:
This change makes it possible to parse composite types (e.g. `list<record<key: string>>`, `record<key: list<int>>`)

### DIFF
--- a/corpus/decl/def.nu
+++ b/corpus/decl/def.nu
@@ -350,3 +350,280 @@ def test [
         (param_type
           (flat_type))))
     (block)))
+
+======
+def-019-type-list-of-list
+======
+
+def test [
+  x: list<list<int>>
+]: nothing -> list<list<int>> {}
+
+-----
+
+(nu_script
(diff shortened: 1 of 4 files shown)


fdncred:
This looks good to me but my only question is will it still parse if it's on different lines. For instance, this is one of the tests.
```nushell
def test [
  x: record<key: list<int>>
]: nothing -> record<key: list<int>> {}
```
but what about
```nushell
def test [x: record<key: list<int>>
]: nothing -> record<key: list<int>> {}
```
or
```nushell
def test [x: record<key: list<int>>]: nothing -> record<key: list<int>> {
}
```
or all on one line
```nushell
def test [x: record<key: list<int>>]: nothing -> record<key: list<int>> {}
```
do all these work too?

kit494way:
Yes, all of those works.
I added tests for those cases.

fdncred:
Thank you! and thanks for your continued work here!

PR #66, opened 2024-01-11
Add symbols to command identifier

### LINKED ISSUE #63: command identifier with `?` is not parsed correctly
スクリーンショット 2024-01-05 112653

kit494way:
Fix #63 

Although there are many available symbols, I added `?` that is actually used [here](https://github.com/nushell/nushell/blob/62272975f2244229099c903ab43322f3db7e8d00/toolkit.nu#L396) and `!` that is likely to be used.


fdncred:
ya, i've seen many people name custom commands with `?` but that's the only symbol i remember, of course, `-` and `_` are used in naming but not usually at the end. thanks for this.

PR #68, opened 2024-01-12
Fix parsing of closure and record

### LINKED ISSUE #16: blocks cannot be used as values
# Description

since we cannot backtrack, it is difficult to tell apart blocks and records. so currently, blocks are not parsed as values

```nu
# this will not parse
ls | each { $in + 4 }

# however this parses
ls | each {|it| $it + 4 }
```

### LINKED ISSUE #65: Some record keys are not parsed correctly
The following keys are not parsed correctly
- unquoted string contains a hyphen
- number
- variable

スクリーンショット 2024-01-09 225901

### LINKED ISSUE #67: Closure parsed as a record
In the following example:
```nu
#!/usr/bin/env nu

let $list = [a b c d]

# Correctly parsed
$list | filter {|x| $x | str starts-with c }

# Still valid but not correctly parsed
$list | filter { str starts-with c }
```

The closure for filter command is incorrectly parsed in the last line. And this leads to an error in the tree.
tree-sitter-nu3

kit494way:
Fix #67 
Fix #16 
Fix #65 

This change makes it possible to parse closure without parameter pipes.
And following types are parsed as record key
- unquoted string contains a hyphen
- number
- variable
- subexpression

### DIFF
--- a/corpus/expr/closure.nu
+++ b/corpus/expr/closure.nu
@@ -1,0 +1,103 @@
+=====
+closure-001-basic
+=====
+
+[1, 2] | each {|x| $x + 1 }
+
+-----
+
+(nu_script
+  (pipeline
+    (pipe_element
+      (val_list
+        (val_number)
+        (val_number)))
+    (pipe_element
+      (command
+        (cmd_identifier)
+        (val_closure
+          (parameter_pipes
+            (parameter
+              (identifier)))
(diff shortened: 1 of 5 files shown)


fdncred:
wow, this is great! thanks for fixing those long-standing issues!

PR #69, opened 2024-01-13
Fix parsing of records when the key is the same as the keyword

kit494way:
After #68 , parsing of record fails when the key is same as the keyword (e.g. `error`).
This pr fixes the bug.

Before
スクリーンショット 2024-01-13 094954

After
スクリーンショット 2024-01-13 095204


fdncred:
nice work!

PR #72, opened 2024-01-21
Fix parsing of units

### LINKED ISSUE #71: `duration` is parsed as a `string` when used as an argument.
`tree-sitter-nu` version: 4e5567e8b36f39b3484e20ad76c2e0149bb07957
ts_2

kit494way:
Fix #71


fdncred:
nice work, thanks again!

PR #73, opened 2024-01-23
Fix parsing of range

### LINKED ISSUE #53: Parsing of range fails
スクリーンショット 2024-01-02 172309

This bug is introduced by this change #50.
It seems that the rule, which parses a number omitted after decimal point (e.g. `1.`) as float, parses head of range value as float incorrectly.

kit494way:
Fix #53 , and make it possible to parse range using variables and subexpressions.

### DIFF
--- a/corpus/expr/range.nu
+++ b/corpus/expr/range.nu
@@ -1,0 +1,813 @@
+=====
+range-001-basic
+=====
+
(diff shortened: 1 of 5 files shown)


fdncred:
nice job! and i love all the tests!

PR #74, opened 2024-01-24
Fix parsing of def using flags

### LINKED ISSUE #49: Parsing of `def` with flags fails
スクリーンショット 2023-12-27 221408

kit494way:
Fix #49

### DIFF
--- a/corpus/decl/def.nu
+++ b/corpus/decl/def.nu
@@ -724,3 +724,29 @@ def a_-!? [] {}
     (cmd_identifier)
     (parameter_bracks)
     (block)))
+
+======
+def-032-flag
+======
+
+def --env foo [] {}
+def --env --wrapped bar [...args] {}
+
+-----
+
+(nu_script
+  (decl_def
+    (long_flag)
+    (cmd_identifier)
+    (parameter_bracks)
+    (block))
+  (decl_def
+    (long_flag)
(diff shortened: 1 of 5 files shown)


fdncred:
I see these tested
```nushell
def --env foo [] {}
def --env --wrapped bar [...args] {}
```
but does it cover these?
```
def --wrapped --env foo [] {}
def --wrapped bar [...args] {}
```

stevenxxiu:
I also wonder if `def-env` should be changed to `def --env`?

fdncred:
@stevenxxiu I don't follow. Where is `def-env` used in this PR?

stevenxxiu:
@fdncred There are some in the tests and some in `grammar.js`, but this is now removed.

fdncred:
if we still have `def-env` anywhere, it needs to be removed, but not necessarily in this PR.

stevenxxiu:
That could make sense, but would it be better to directly change the `def-env` tests to `def --env`, and remove `def-env` at the same time?

kit494way:
@fdncred @stevenxxiu Thanks for your reviews. I will remove `def-env` and fix the tests in this PR.

kit494way:
I removed `def-env` and fixed the tests.

fdncred:
Thanks. looks good!

PR #78, opened 2024-01-29
Fix parsing unquoted

### LINKED ISSUE #77: `inf` is parsed as a number unless quoted
`tree-sitter-nu` version: 2d0dd587dbfc3363d2af4e4141833e718647a67e
Screenshot_20240125_140016

kit494way:
Fix #77 
Fix parsing of special keywords (e.g. `inf`, `nan`) to ignore case.
Add a special keyword `infinity`.
Remove the unused keyword `def-env` that was not removed in the #74 pull request.
Improve parsing of unquoted string.

### DIFF
--- a/corpus/expr/list.nu
+++ b/corpus/expr/list.nu
@@ -25,3 +25,197 @@ list-001-item-string
         (val_string)
         (val_string)
         (val_string)))))
+
+=====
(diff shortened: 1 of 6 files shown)


fdncred:
wow, thanks for this. looks like a pain to implement. 😆

PR #79, opened 2024-02-02
Fix parsing of match pattern

### LINKED ISSUE #60: Parsing of list destruction match pattern fails
1. fails to parse with guard
2. parsed as range value incorrectly without comma

スクリーンショット 2024-01-04 130054

スクリーンショット 2024-01-04 132022

kit494way:
Fix #60 

This PR changes the syntax tree of list destrcture pattern.

Before this change, only `[$x ..$y]` is supported as list destrcture pattern.
However, there are various cases (e.g. `[1 ..$x]`, [$x $y ..]).
To support other various cases, this PR changes the way of parsing list destrcture pattern.

Before this change, parsing list destrcture pattern generates different syntax tree from list pattern.
For example, when parsing `match $xs {[$x ..$y] => {}}`, `(val_list)` and `(val_variable)` nodes were not generated under `(match_pattern)` node before this change as follows.

```
(nu_script
  (pipeline
    (pipe_element
      (ctrl_match
        (val_variable
          (identifier))
        (match_arm
          (match_pattern
            (identifier) ; <-- (val_list) and (val_variable) are not generated
            (identifier))
          (block))))))
```

After this change, the parsing is done without distinguishing between list destrcture patten and list pattern.
`(val_list)` and `(val_variable)` nodes are also generated when parsing list destrcture pattern as follows.

```
(nu_script
  (pipeline
    (pipe_element
      (ctrl_match
        (val_variable
          (identifier))
        (match_arm
          (match_pattern
            (val_list
              (val_variable
                (identifier))
              (val_variable
                (identifier))))
          (block))))))
```

### DIFF
--- a/corpus/ctrl/match.nu
+++ b/corpus/ctrl/match.nu
@@ -65,3 +65,192 @@ match $x {
             (pipeline
               (pipe_element
                 (val_record)))))))))
+
+=====
+match-003-list
+=====
+
+match $xs {
+   [] => {},
+   [1] => {},
+   [1, 2] => {},
+   [$x] => {},
+   _ => {}
+}
+
+-----
+
+(nu_script
+  (pipeline
+    (pipe_element
+      (ctrl_match
+        (val_variable
+          (identifier))
+        (match_arm
+          (match_pattern
+            (val_list))
+          (block))
+        (match_arm
+          (match_pattern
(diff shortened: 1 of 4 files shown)


fdncred:
wow! nice work, thanks!

PR #81, opened 2024-02-08
Fix pasing of match pattern

kit494way:
Fix to correctly parse following two match patterns
- a short-hand of record pattern (e.g. `{$key}`)
- unquoted string


fdncred:
Looks like the ci isn't happy. I was thinking prettier ran automatically but I guess not?

kit494way:
prettier runs automatically, but only checks.

fdncred:
great! thanks for fixing it.

PR #82, opened 2024-03-07
Add treesitter indent rules

zDonik1:
I am pretty new to nu-lang, so the indent rules might not be exhaustive, but I decided to add it here since it was missing.

### DIFF
--- a/queries/nu/indents.scm
+++ b/queries/nu/indents.scm
@@ -1,0 +1,24 @@
+[
+  (pipeline)
+  (pipe_element)
+  (expr_parenthesized)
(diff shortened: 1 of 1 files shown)


fdncred:
I thought we had indent rules before but maybe I'm just remembering wrong?

zDonik1:
It looks like it didn't have one. I will add indenting for parenthesis expressions as well.

zDonik1:
I have added some more rules but won't be working on it anymore for now.

fdncred:
Thanks!

PR #87, opened 2024-04-20
Fix parsing of where

### LINKED ISSUE #32: Parser breaks with this syntax
With the `type@'type'` on the command parameter, the syntax highlighting breaks.

Example of the code:
```nu
export def-env main [
	command?: string@'nu-complete rbenv',
	...args
] {
    let new_env = if $command in ["rehash", "shell"] {
        # implement each on indiviudaly e.g.
        if $command == "shell" {
            { RBENV_VERSION_OLD: $env.RBENV_VERSION RBENV_VERSION: $args.0 }
        } else {
            error make { msg: $"`($command)` command is not supported yet" }
        }
    } else {
    	if ($command | is-empty) {
    		^rbenv
    	} else {
        	^rbenv $command $args
    	}
        {}
    }
    load-env $new_env
}

def 'nu-complete rbenv' [] {
    ^rbenv help
    | lines
    | where ($it | str starts-with "   ")
    | each {|entry| $entry | split row ' ' | get 0 }
}
```

Screenshoot:
Screenshot 2023-09-24 at 23 38 56

kit494way:
Fix #32

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -686,7 +686,15 @@ module.exports = grammar({
 
     // the where command has a unique argument pattern that breaks the
     // general command parsing, so we handle it separately
-    where_command: ($) => seq("where", $._where_predicate),
+    where_command: ($) =>
+      seq(
+        "where",
+        choice(
+          $._where_predicate,
+          $._expression,
+          alias($.block, $.val_closure),
+        ),
+      ),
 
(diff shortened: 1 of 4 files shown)


fdncred:
Thank you so much for your continued help here!

PR #90, opened 2024-05-23
Fixes to parsing of use, else & commands

### LINKED ISSUE #83: script of the sort "^'bin' command" is not parsed
Just as the title says, having `^'bin' command` in the nu file doesn't get parsed.

image

### LINKED ISSUE #88: Wrong highlighting in if else statement.
Treesitter highlighting fails when else flag is missing a closure ``else $statement``.

**Look at the let keywords below.**
Example without closure:
image
Example with closure:
image

Tested in neovim.

### LINKED ISSUE #89: use *file* identifier: produces a tree-sitter error
Using the pattern ``use file identifier`` produces a TS error. This happens when the identifier is not put in a list ex: ``use file [identifier`` - this produces no error. Also happens when * is used instead of an identifier.

Picture (look at the highlighting of the use - on the line before the incorrectly highlighted use there is an TS error, which can be seen via the TSPlayground):
image

BigCrow:
### Description
Fix parsing of **use**, **else** & **commands**, by taking into account additional valid syntax.

### Issues:
fixes #83 
fixes #88 
fixes #89 

#### command
Split the command rule to accomodate 2 more syntaxes: ``^'ls'`` and ``^()``.

#### else
Allow non-block syntax for else (in comparison with if, which can't take anything, but a block).

#### use
Allow use to take pipe input.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -345,7 +345,7 @@ module.exports = grammar({
           seq(
             KEYWORD().else,
             choice(
-              field("else_block", $.block),
+              field("else_block", choice($.block, $._expression, $.cmd_identifier)),
               field("else_branch", $.ctrl_if),
             ),
           ),
@@ -363,7 +363,7 @@ module.exports = grammar({
               optional("\n"),
               KEYWORD().else,
               choice(
-                field("else_block", $.block),
+                field("else_block", choice($.block, $._expression, $.cmd_identifier)),
                 field("else_branch", alias($.ctrl_if_parenthesized, $.ctrl_if)),
               ),
             ),
--- a/grammar.js
+++ b/grammar.js
@@ -118,7 +118,7 @@ module.exports = grammar({
         seq(
           optional(MODIFIER().visibility),
           KEYWORD().use,
-          field("module", choice($.val_string, $.unquoted)),
+          field("module", choice($.val_string, $.expr_parenthesized, $.unquoted)),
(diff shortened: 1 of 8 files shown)


fdncred:
looks like we require prettier to be run - we might should say that somewhere.

BigCrow:
Formatted with prettier.

BigCrow:
Update the use rules to account for interpolated string as suggested by [kubouch](https://github.com/nushell/tree-sitter-nu/issues/89#issuecomment-2128673923).

BigCrow:
Note that I didn't regenerate for the last change, in order to keep the diff minimal.
After mergin the project **should be regenerated** with ``tree-sitter generate`` and the 3 generated src files should be commited for the project to work properly with python.

BigCrow:
> Note that I didn't regenerate for the last change, in order to keep the diff minimal. After mergin the project **should be regenerated** with `tree-sitter generate` and the 3 generated src files should be commited for the project to work properly with python.

Actually not a good idea - the github tests require the grammar to already be generated. I regenerated it so it could pass the new tests.

fdncred:
I updated the packages with `npm update`, `npm outdated`, and `npm install` and I got the windows stuff to work, at least on my windows system. let's see if ci passes.

fdncred:
yay, it worked. let's move forward! Thanks @BigCrow!

PR #92, opened 2024-05-24
Source issue

CabalCrow:
### Description
Fix parsing of **source**, by taking into account additional syntax.

### Bug:
Creates an error in TS. Does not affect current highlighting, but if someone wants to use the source argument for highlighting it wouldn't be found when the error is triggered.

#### source
Allow quoted & interpolated strings as well as pipes in parenthesis as arguments to source. Also updates tests for source for the new grammar.

### DIFF
--- a/src/grammar.json
+++ b/src/grammar.json
@@ -3999,7 +3999,19 @@
               },
               {
                 "type": "SYMBOL",
+                "name": "val_string"
+              },
+              {
+                "type": "SYMBOL",
+                "name": "expr_parenthesized"
+              },
+              {
+                "type": "SYMBOL",
                 "name": "val_variable"
+              },
+              {
+                "type": "SYMBOL",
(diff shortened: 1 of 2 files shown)


fdncred:
> Creates an error in TS. 

what does that mean?

BigCrow:
> > Creates an error in TS.
> 
> what does that mean?

Creates an ERROR query in tree-sitter. This means there is undefined behaviour to the rules & TS just gaves up.

The problem essentially is that it doesn't parse the whole ``source something`` structure correctly. What happens is: it parses the ``source`` as a keyword & highlight it as such, it parses the ``something`` as whatever it is & highlight it as such. However it doesn't actually create the whole ``source`` structure, so if someone wants to create specific highlighting that is specific to the ``source file`` group they won't be able to.

This fixes this by adding a more syntax structures to the ``source`` rules.

PR #94, opened 2024-05-26
Fix parsing of record & list bodies.

BigCrow:
### Brief Description
Change the node structure of the record & list bodies to better represent the actual syntax of the language. This fixes some bugs as well as make the syntax more consistent & easier to read.

### Fixes
- TS can't properly detect durantion values in records. It is instead deteced as integer & a command node that create an error:
```nu
{
    key: 5sec
}
```
- TS can't properly detect when value & key begins & allow invald syntax with value & key having no space or any separator in between. The following example is considered valid by TS:
```nu
{
    key: 1536another_key: 8549
}
```

### What is changed
- Records use a new body structure which manages entries & separators in between. This fixes the separator bugs described above & creates new nodes that can be used for highlighting & indenting.
- Lists also have their structure changed in the same manner (both records & list body acts the same in nushell syntax, with exception to what is actually a record entry ``key: value`` vs a list entry ``value``.
- Record tests are added for the bugs described above. All tests have been updated with the new node syntax.

### Details
- New function is added to grammar js for general body structure. It consist of repeated statement all seperated by a mandotory separator, followed by a final statement with an optional separator.
- This function is applied to the bodies of lists & records with the corresponding entries.
- List has an added node for the list_entry itself (before it was just hardcoded into the list rule)
- Separator is added for entries ``[,\s\n\r]``
- Comment is added on top of the code to improve LSP parsing of code with tsserver (commonly used by other TS projects for grammar.sj)
- Tests are updated with new new node syntax - I've individually checked the diffs of all tests to make sure they are working as indented
- New test has been added to check if records can accept duration input.
- New test has been added to test if records accept invaild input with no separator between value & key.

### DIFF
--- a/test/corpus/expr/closure.nu
+++ b/test/corpus/expr/closure.nu
@@ -1,4 +1,163 @@
 =====
+closure-001-basic
+=====
+
+[1, 2] | each {|x| $x + 1 }
+
+-----
+
+(nu_script
+  (pipeline
+    (pipe_element
+      (val_list
+        (list_body
+          (val_entry
+            (val_number))
+          (val_entry
+            (val_number)))))
+    (pipe_element
+      (command
+        (cmd_identifier)
+        (val_closure
+          (parameter_pipes
+            (parameter
+              (identifier)))
+          (pipeline
+            (pipe_element
+              (expr_binary
+                (val_variable
+                  (identifier))
+                (val_number)))))))))
+
+=====
+closure-002-closure-without-parameter
+=====
+
+[1, 2, 3] | each {|| $in + 2 }
+
+-----
+
+(nu_script
+  (pipeline
+    (pipe_element
+      (val_list
+        (list_body
+          (val_entry
+            (val_number))
+          (val_entry
+            (val_number))
+          (val_entry
+            (val_number)))))
+    (pipe_element
+      (command
+        (cmd_identifier)
+        (val_closure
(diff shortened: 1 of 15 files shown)


fdncred (commented):
I think this mostly looks good but what I don't understand is why the closure-xxx-yyyy tests were removed, except for 007. I expected them to be changed but not removed. Same for export-env-001-smoke-test.

BigCrow:
It must have been a mistake on my part. Thanks for catching that - I re added the lost tests.

test/corpus/expr/closure.nu
```diff
@@ -163,49 +172,22 @@ closure-007-record-value
...
-                  (val_number))))))
-        (record_entry
-          (identifier)
-          (val_closure
-            (pipeline
-              (pipe_element
-                (val_string)))))))))
-
-=====
-closure-008-while
```
fdncred, line 186: was closure-008-while just left out because it's a dupe or was this accidental?

BigCrow: Accidental thanks - I will go through the diff once again to make sure I haven't forgotten anything. But none of the test should be removed, they should only get changes.

BigCrow: I think everything should be fine now - I went through the test diffs & I don't see anymore lost tests.

fdncred:
Thanks

PR #95, opened 2024-05-27
Warning & error suppression for LSP

CabalCrow:
This adds several warning & error suppressions for the tsserver lsp. Doesn't actually change any functionality (if you tree-sitter generate you will see that nothing would get changed).

- add type docs to functions
- supress tsserver errors (they are not actual errors, tsserver just can't properly diagnose the grammar.js)
- add _ to unused arguments

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -38,13 +38,22 @@ module.exports = grammar({
 
     shebang: (_$) => seq("#!", /.*\n/),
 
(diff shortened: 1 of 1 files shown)


fdncred:
I don't have a problem with this PR but what is `tsserver`?

fdncred:
ok, that makes sense that you'd use the typescript lsp for grammar.js. seem benign enough.

PR #96, opened 2024-05-27
Quickfix for separator token using non-lexical precedence.

CabalCrow:
The separator node introduced in #94 should use lexical precedence instead of parse precedence. This is a quick fix to that.
Currently I don't know any issues introduced by this, but this could potentially create an issue.


fdncred:
Thanks

PR #99, opened 2024-06-12
nvim queries: Conform to commonly used highlight names

hedyhli:
This updates:

| Query | Highlight name | Like in... |
|-|-|-|
| Punctuation `:`, `,` | `punctuation.special` | TOML, Bash, JSON, ... |
| Key fields in tables/records | `variable.member` | Go, Zig, V, TOML |
| Returns `->` | `punctuation.special` | Rust |
| Variable `$` | `punctuation.special` | Bash, Awk |
| Variable identifier | `variable` | Bash, Awk, Perl, Fish |
| Shebang | `keyword.directive` | Bash, Awk |
| Built-in types | `type` | Zig et al[^1] |
| Numbers | `number` | Zig, Rust, Python, TOML, Bash, ... |
| Arguments | `variable.parameter` |  Bash[^2] |
| Redirect operator | `operator` | Bash, Fish |
| Redirect destination | `variable.parameter` | Bash |

[^1]: Note that for built-in types, languages like Rust, Python, Go differentiates highlights of built-in types and other types (such as structs), but there are no other types in Nu, I'll leave this up to debate, IMO a generic type will suffice here.

[^2]: Perl, Fish and Awk all leave these unhighlighted.

Also adds highlights for documentation comments for functions and parameters.

I've tried to put newly introduced highlight queries into the correct groups in the file, let me know if there are any other prevailing styles I should follow!

### DIFF
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -56,7 +56,7 @@
 
 ;;; ---
 ;;; literals
-(val_number) @constant.numeric
+(val_number) @number
 (val_duration unit: _ @variable.parameter)
 (val_filesize unit: _ @variable.parameter)
 (val_binary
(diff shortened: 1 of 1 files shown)


fdncred:
I'm willing to give this a try. I like this idea for sure!! Thanks

fdncred:
How do you control which highlights.scm file nvim uses for a particular language?

hedyhli:
> How do you control which highlights.scm file nvim uses for a particular language?

If the Lua plugin in this repo is used, the file will be in `queries/nu/` in the runtime. And nvim-treesitter will see the filetype, nu, then fetch `queries/<nu>/` in the runtime. If there are multiple, precendence is determined in a similar manner to `$PATH`, as listed in `set rtp?`.

fdncred:
That's very helpful. Thanks

fdncred:
It looks like the highlighting is a bit off now. This is from the toolkit.nu file in the root of the nushell repo.

## Before
image
## After
image

fdncred:
@hedyhli wondering about your thoughts on this broken highlighting?

fdncred:
@hedyhli will you have an opportunity to look at [this issue](https://github.com/nushell/tree-sitter-nu/pull/99#issuecomment-2165419934)?

PR #100, opened 2024-06-12
nvim queries: Highlight built-in commands

hedyhli:
I thought it'd be great to highlight built-in commands differently to custom commands similar to what Nu does in the prompt line. It should be fairly easy to use the oneliner included in the comments on later versions of Nu to update the list.

Reference:
- Bash: https://github.com/nvim-treesitter/nvim-treesitter/blob/b694628cbe513ae8bcf9fb5ac791a73ca3fd5320/runtime/queries/bash/highlights.scm#L169-L176
- Fish: https://github.com/nvim-treesitter/nvim-treesitter/blob/b694628cbe513ae8bcf9fb5ac791a73ca3fd5320/runtime/queries/fish/highlights.scm#L115-L124

Core built-ins are filtered out because they are already specified higher up in the file as keywords.

All commands that aren't in the list are still highlighted with `@function`

### DIFF
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -221,13 +221,40 @@
 (scope_pattern [(wild_card) @function])
 
 (cmd_identifier) @function
+; generated with Nu 0.93.0
+; > help commands
+;   | filter { $in.command_type == builtin and $in.category != core }
+;   | each {$'"($in.name | split row " " | $in.0)"'}
+;   | uniq
+;   | str join ' '
+(command
+  head: [
+    (cmd_identifier) @function.builtin
+    (#any-of? @function.builtin
+     "all" "ansi" "any" "append" "ast" "bits" "bytes" "cal" "cd" "char" "clear"
+     "collect" "columns" "compact" "complete" "config" "cp" "date" "debug"
+     "decode" "default" "detect" "dfr" "drop" "du" "each" "encode" "enumerate"
+     "every" "exec" "exit" "explain" "explore" "export-env" "fill" "filter"
+     "find" "first" "flatten" "fmt" "format" "from" "generate" "get" "glob"
+     "grid" "group" "group-by" "hash" "headers" "histogram" "history" "http"
+     "input" "insert" "inspect" "interleave" "into" "is-empty" "is-not-empty"
+     "is-terminal" "items" "join" "keybindings" "kill" "last" "length"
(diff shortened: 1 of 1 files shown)


fdncred:
I like the idea, but it seems like there's a lot of commands missing like the commands with spaces in the name, also plugins and ones like registry query that only exist on Windows. I think if we can get a complete list, this will be a nice add.  I have to do the same thing for the vscode extension so I may have a script laying around somewhere.

fdncred:
With the vscode-extension I just do this on windows and mac/linux, combine the files and sort uniq them for the complete list. You also have to have the plugins installed and registered when doing this.
```nushell
scope commands | where type == built-in | get name | to text | save mac-cmds.txt
```

hedyhli:
> I like the idea, but it seems like there's a lot of commands missing like the commands with spaces in the name

I didn't see an easy way to include them because subcommands are parsed as arguments and not part of the command itself. If we match the command and subcommand separately there will be incorrect highlighting (eg `my-command split` vs `str split`).

> also plugins (unless we don't want them)

I'm not sure if that's a good idea, wouldn't that be different for everyone? The highlight file will have to be the same.

> and ones like registry query that only exist on Windows

Ah, good point!

> I'm also not sure why core commands are excluded.

They are already specified in the highlight file and highlighted as keywords, which I've explained in the PR description :)

> Core built-ins are filtered out because they are already specified higher up in the file as keywords.

fdncred:
ok, let's move forward with this then. thanks!

hedyhli:
I've got an idea how built-in subcommands can also be highlighted, I'll PR if it works -- indeed, for completeness highlighting both words would be the ideal situation.

hedyhli:
@fdncred I've gotten in to work, but it's a little verbose -- though the simplest I could find:



Here, `str split` and `str another` are the only two valid built-in commands, where `function.builtin` is orange and `function` is blue.

The `.subcommand` part is only for descriptive purposes in the `highlight.scm` file, because the two groups must be different for the comparison to work properly. It makes use of Neovim's fallback mechanism, where if a highlight group `A.B.C` is not found, it tries `A.B`, then `A` until a color is found.

With this method, all subcommands for all built-in commands will have to be repeated like the first two lines above.

What do we think?

fdncred:
> What do we think?

"we" like it. I'd go for it because it's better than what we had before.

PR #101, opened 2024-06-13
Fix highlight of `where` command as builtin

hedyhli:
Apologies, in #100 I hadn't realized the where command is special-cased in the parser and hence requires a separate capture to highlight properly. This ensures the `where` command is also highlighted as built-in.


fdncred:
thanks

Issue #103, opened 2024-08-19
Publish SemVer tags to luarocks.org?


mrcjkb:
Hey :wave: 

[rocks.nvim supports using tree-sitter parsers that have been published to luarocks.org as plugin dependencies](https://github.com/nvim-neorocks/rocks.nvim?tab=readme-ov-file#deciduous_tree-enhanced-tree-sitter-support).

This means that rocks.nvim users can just `:Rocks install my-nushell-plugin.nvim`, and the tree-sitter parser will be installed automatically, without having to use nvim-treesitter to install parsers.

For parsers that are supported by nvim-treesitter, we have an automated workflow to package/publish them using [the NURR](https://github.com/nvim-neorocks/nurr).
But for parsers that aren't natively supported by nvim-treesitter, it's better if the authors manage them themselves.

Some examples:

- [tree-sitter-norg](https://luarocks.org/modules/neorg/tree-sitter-norg)
- [tree-sitter-orgmode](https://luarocks.org/modules/kristijanhusak/tree-sitter-orgmode).

Would you be open to publishing SemVer releases (e.g. using the release-please github action) and pubishing `tree-sitter-nu` to luarocks.org?
If so, I'd be happy to set up the GitHub actions for you.

fdncred:
👋🏻 

We're of two minds in regard to publishing.
1. We're happy to have tree-sitter-nu in as many places as possible for users to easily install and use it.
2. It ends up being a big pain to support because people change things and then publishing breaks without anyone who cares to support it.

mrcjkb:
Thanks for the feedback. 
In that case I suppose it's better for you if we add it to the NURR.

mrcjkb:
I have [added tree-sitter-nu to the NURR](https://github.com/nvim-neorocks/nurr/commit/e1a3e5f6ff87c5f502210551193c1c0a813432b4), so a `scm` version is [now available on luarocks.org](https://luarocks.org/modules/neorocks/tree-sitter-nu).

It would be nice to have semver releases so that users can pin versions.
Are there any plans for that?

fdncred:
Yes, but probably not until we have a maintainer passionate enough about tree-sitter-nu to finish the grammar and highlighting and do releases.

PR #104, opened 2024-08-30
fix: some fixes for use with Neovim

mrcjkb:
This PR adds the following fixes for use with Neovim:

- Make sure the `plugin` script can only be sourced once (this is a best practise for plugins).
- Move the registration of the `nu` filetype to an `ftdetect/nu.lua` script, as is common practise for Neovim plugins.
- Move setting the commentstring for `nu` files to `ftplugin/nu.lua` (so there's no need to create an autocommand).
- Make sure the plugin doesn't error if nvim-treesitter isn't installed. This is in preparation for packaging this parser for luarocks.org, so that rocks.nvim/luarocks users can use it without having to install nvim-treesitter.
  See also: #103

### DIFF
--- a/ftdetect/nu.lua
+++ b/ftdetect/nu.lua
@@ -1,0 +1,1 @@
+vim.filetype.add({ extension = { nu = "nu" } })
--- a/ftplugin/nu.lua
+++ b/ftplugin/nu.lua
@@ -1,0 +1,2 @@
+local bufnr = vim.api.nvim_get_current_buf()
+vim.bo[bufnr].commentstring = "# %s"
--- a/plugin/init.lua
+++ b/plugin/init.lua
@@ -1,15 +1,15 @@
-vim.filetype.add({ extension = { nu = "nu" } })
+if vim.g.loaded_tree_sitter_nu then
+  return
+end
+vim.g.loaded_tree_sitter_nu = true
 
-vim.api.nvim_create_autocmd("FileType", {
-  pattern = "nu",
-  callback = function(event) vim.bo[event.buf].commentstring = "# %s" end,
-})
-
-require("nvim-treesitter.parsers").get_parser_configs().nu = {
-  install_info = {
-    url = "https://github.com/nushell/tree-sitter-nu",
-    files = { "src/parser.c" },
-    branch = "main",
-  },
-  filetype = "nu",
-}
+pcall(function()
+  require("nvim-treesitter.parsers").get_parser_configs().nu = {
+    install_info = {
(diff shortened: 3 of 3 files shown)


fdncred:
Thanks for trying to make this better. However, I'm not sure I'm a fan of this PR since it makes it more complicated for people to understand what to do. I use nvim and don't have ftdetect or ftplugin at all so it makes me think that this is for some specific configuration that the average nvim user may not have.

I'm fine with fixes though if there are problems with the script. Let's see what others say.

mrcjkb:
[`:h ftdetect`](https://neovim.io/doc/user/filetype.html#ftdetect) and [`:h ftplugin`](https://neovim.io/doc/user/usr_41.html#ftplugin) are built-in Neovim features that are enabled by default. 

There's nothing users need to do for this to work. Neovim sources the scripts automatically at the appropriate times (as opposed to sourcing everything at startup, regardless of whether a nushell file has been opened or not, which is what the current implementation does).

fdncred:
I'm also wondering if this stuff needs to go in our integrations repo vs here with a readme linking to the integrations repo.

mrcjkb:
That sounds reasonable to me.

fdncred:
I'd accept it in integrations if
1. you make a couple folders to leave the original one file implementation
2. you make a new folder for these changes with a readme that explains why it exists

Then you could come back here and add a readme.md to the plugin folder that points there.

should we move this one file implementation over there in yet another folder?

mrcjkb:
I don't understand your suggestion. I don't use this parser myself, I'm just responding to [a request to get it packaged](https://github.com/nvim-neorocks/rocks.nvim/issues/111#issuecomment-2289390421).

Come to think of it, I believe the proper way to go for filetype detection is to upstream the `.nu` filetype to vim (which would result in it being backported to Neovim eventually).

After that, the `ftdetect` and `ftplugin` scripts would be redundant.
Registering with nvim-treesitter isn't needed for rocks.nvim users, hence the `pcall` as they likely won't have nvim-treesitter installed. But I can just exclude the `plugin` script from the luarocks package.

mrcjkb:
the commentstring patch [has been backported to Neovim](https://github.com/neovim/neovim/pull/30214).

Closing, as I can just exclude the `plugin` directory in the luarocks package.

My recommendation would be for you to remove the redundant filetype registration, to only set the commentstring if `vim.fn.has('nvim-0.11') ~= 1` and to wrap the nvim-treesitter usage in a `pcall`.

PR #106, opened 2024-09-17
feat: detect nurfiles as nu and add a sample nurfile

mrdgo:
nur is a job runner, like just but on nu


fdncred:
seems good to me. thanks!

Issue #107, opened 2024-09-18
Error on opening *.nu files


PAF13:
after adding the Lazy dependencies to my "nvim-treesitter/nvim-treesitter"  script i get the error below.

I am not sure where to start looking to find the problem. My first guess is that it is compiled with zig but i cant get the other compilers to work right now.

`Error in decoration provider treesitter/highlighter.win:
Error executing lua: .../current/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 221:19. Invalid node type "long_flag_identifier":         
(param_long_flag (long_flag_identifier) @variable.parameter)
stack traceback:
    [C]: in function '_ts_parse_query'
    .../current/share/nvim/runtime/lua/vim/treesitter/query.lua:252: in function 'fn'
    ...vim/current/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'fn'
    ...vim/current/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'get'
    ...nt/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:28: in function 'new'
    ...nt/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:240: in function 'get_query'
    ...nt/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:188: in function 'fn'
    ...t/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:491: in function 'for_each_tree'
    ...nt/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:175: in function 'prep `

fdncred:
It looks like a [recent PR](https://github.com/nushell/tree-sitter-nu/pull/105) broke the highlights.scm. @mrdgo Do you have time to look into this?

mrdgo:
Interesting. Sure, I will look into this!

PAF13:
I dont know yet how treesitter works so I just used `TSInstall nu` and did not use `TSUpdate`. It works now, thank you.

mrdgo:
Glad to help!

Then please also update your config according to #108

fdncred:
But the problem is that the highlights.scm file doesn't work with the current syntax. `TSInstall nu` must be using an old ... something?

mrdgo:
No, I updated it accordingly. If you look at it, there are occurrences of `long_flag_identifier` in lines 221 and 225 :)

fdncred:
[...]

PR #108, opened 2024-09-19
fix: instructions to install the plugin

mrdgo:
Relates to #107

Technically, this doesn't have to happen on every update to the plugin. But I don't think that lazy has a mechanism to detect only changes to `grammar.js` and `queries/nu/*.scm`

Edit: according to OP of #107, this is the fix.


fdncred:
ok, thanks

PR #109, opened 2024-09-19
fix: highlight newly created variables as variables

mrdgo:


fdncred:
nice! I assume this need a :TSUpdate to apply this change, correct?

mrdgo:
> nice! I assume this need a :TSUpdate to apply this change, correct?

Nope, this one doesn't :) Because I only updated the highlights. `TSUpdate` is only required when `grammar.js` updates.

PR #110, opened 2024-09-19
feat: improve distinction of flags and parameters

mrdgo:
This PR does three things:
1. highlight `=` in `--long-flag=value` flags (orange in my case)
2. highlight flag identifiers as parameters (blue)
3. highlight string values as strings (green)

Before:
image

After: 
image

Do we like these changes?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ target/
 
 /build/
 
+# C lsp cache
+.ccls-cache/
+
 # Workspace Config
 .nvim/
 # .vscode/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,9 +51,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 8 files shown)


fdncred:
I'm not a fan of everything being the same color. I kind of like the parameters being a different color from the commands. I do like the `--time-style=long-iso` color distinction. It's hard to tell on your theme, maybe your commands are yellow and the parameters are dark yellow?

mrdgo:
The command is bold and green while strings are italic and green.

I would also be happy giving flag identifiers the `type` query:
image

fdncred:
I like this most recent screenshot better. Lemme know when you're ready to land.

mrdgo:
Last commit made exactly that happen :) From my side, this can be merged

fdncred:
Thanks!

PR #112, opened 2024-09-20
fix: flags again

mrdgo:
Last time, I changed identifiers to @type. Most of the time, this is
fine. But when defining a new command with flag parameters, these can
have types and the same highlighting. So now they are attributes and we
have color distinction in both cases.

Before:
image
image

After:
image
image

Since we currently don't have attributes (usually annotations), this is a unique hl-group for flag identifiers.


fdncred:
nice, i like the way that looks. thanks!

Issue #97, opened 2024-05-29
Add support for new nushell feature - raw strings


CabalCrow:
https://www.nushell.sh/blog/2024-05-28-nushell_0_94_0.html#new-language-feature-raw-strings-toc

mrdgo:
I'll take a look. This will probably require a [custom scanner](https://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners) - so it'll likely take a bit longer  to pull off.

fdncred:
closed by https://github.com/nushell/tree-sitter-nu/pull/113

PR #113, opened 2024-09-20
feat: introduce raw strings

mrdgo:
Support for raw strings. The problem was leading whitespace :roll_eyes: 

image

Edit: does anyone have Windows knowledge?


mrdgo:
Implementation  specific notes:

- in grammar.js, there is only an additional field in the choice() of string values
- the actual magic happens in `src/scanner.c`
- We have to manually count the number of `#` symbols on both sides
- Currently, I can use this implementation to successfully `tree-sitter parse` a file containing only `r#'string'#`
- However, the tests fail - in this scenario, it detects a command and a comment
- I feel like the parser behaves differently between `tree-sitter parse` and `tree-sitter test`. This is the current challenge

NotTheDr01ds:
Does anything in the [Rust tree-sitter grammar](https://github.com/search?q=repo%3Atree-sitter%2Ftree-sitter-rust%20%22r%23%22&type=code) help?

fdncred:
Good point. Rust definitely has very similar raw strings and they have a custom scanner.c too.
https://github.com/tree-sitter/tree-sitter-rust/blob/6b7d1fc73ded57f73b1619bcf4371618212208b1/src/scanner.c#L8-L10

suimong:
And there's also Nickel who use a quite similar [syntax for multiline string](https://nickel-lang.org/user-manual/syntax#strings) (`m%...%"<string>"%...%`), and a similarly implemented `scanner.c`: https://github.com/nickel-lang/tree-sitter-nickel/blob/88d836a24b3b11c8720874a1a9286b8ae838d30a/src/scanner.c#L178

mrdgo:
Thanks for all the suggestions! ♥️

When I find the time, I'll look into those scanners.

mrdgo:
Sadly, the code in those repos code does the same things that I do. Will need to dig deeper.

mrdgo:
Any ideas why this
```js
    _raw_str: ($) => seq("r", $._nested_hash_escaped_str),
    _nested_hash_escaped_str: ($) => seq("#", choice($._inner_raw_string, $._nested_hash_escaped_str), "#"),
    _inner_raw_string: ($) => seq('"', $._raw_str_content, '"'),
    _raw_str_content: (_$) => /.*/,
```
wouldn't cut it? I'd really prefer a recursive definition

mrdgo:
Done, this can be merged. I will add another PR directly afterwards with an optional  extension that we  can discuss seperately.

fdncred:
i wonder what's going on with the windos build? 🤔

mrdgo:
> i wonder what's going on with the windos build? 🤔 

I'll take a look later, maybe I have to declare `scanner.c` somewhere in the CI spec as well

fdncred:
@mrdgo I can try and help on Windows. What do I need to run? I just did this.
```shell
~\source\repos\forks\tree-sitter-nu> [raw-strings +1 ~0 -0 !]
❯ npx tree-sitter test
  comment:
    comment:
```

fdncred:
oof, my tree-sitter-cli was out of date. i couldn't even get `tree-sitter parse test.nu` to work until I updated it. Now I can get output from `tree-sitter test` and `npx tree-sitter test`.
```shell
❯ npx tree-sitter test                                                                                                                                                          09:53:13 AM
  comment:
    comment:
        1. ✓ comment-001-single
        2. ✓ comment-002-before-command
        3. ✓ comment-003-empty-before-command
        4. ✓ comment-003-end-of-line
        5. ✓ comment-003-empty-end-of-line
        6. ✓ comment-003-empty-end-of-line-before-command
  ctrl:
    else:
        7. ✓ else-001-block
        8. ✓ else-002-command
        9. ✓ else-003-command
       10. ✓ else-004-command

... and so on
```

I also updated all the dependencies in the package.json. I can try to push these changes.

fdncred:
ya, ci passes!

scanner.obj
mrdgo, line 1: I don't think that this file should be under version control

fdncred:
i can't really tell what you changed due to the force push.

.github/workflows/main.yml
```diff
@@ -29,4 +29,4 @@ jobs:
         run: npx prettier . --check
 
       - name: test
-        run: npx tree-sitter test
+        run: npx tree-sitter test --debug
```
fdncred, line 32: not sure --debug is needed here any longer

mrdgo: of course not, I was just about to remove that :)

mrdgo:
> i can't really tell what you changed due to the force push.

I removed `scanner.obj`, you probably accidentally committed a compiled artifact

fdncred:
looks like we're ready to go now. one last check, ready to land this?

mrdgo:
I checked for stale comments and debug artifacts. And I think, it's ready to be merged :)

(Btw. you will have to update the parser config, i.e., add `src/scanner.c`, additionally to the regular steps)

mrdgo:
> > (Btw. you will have to update the parser config, i.e., add src/scanner.c, additionally to the regular steps)
> 
> I'm not sure what this means. Where would this need to be updated?

Please take a look at the updated installation guide. We now do this:
```lua
require("nvim-treesitter.parsers").get_parser_configs().nu = {
	install_info = {
		url = "https://github.com/nushell/tree-sitter-nu",
		files = { "src/parser.c", "src/scanner.c" },  -- "src/scanner.c" is new
		branch = "main",
	},
	filetype = "nu",
}
```

fdncred:
gotcha, thanks!

mrdgo:
Closes #97

fdncred:
wow, added that snippet to my init.lua and did `:TSUpdate nu`. It built and compiled fine. Start nvim with toolkit.nu and nvim exits with
```nushell
❯ $env.LAST_EXIT_CODE
-1073741819
```

mrdgo:
> wow, added that snippet to my init.lua and did `:TSUpdate nu`. It built and compiled fine. Start nvim with toolkit.nu and nvim exits with
> 
> ```nushell
> ❯ $env.LAST_EXIT_CODE
> -1073741819
> ```

Oof, that's unintended. Sadly, I can't reproduce. Do you have your config on github? On my setup, nvim sometimes crashes when I try to open it with a file - `nvim $FILE` crashes, `nvim` starts and can open the file afterwards.

fdncred:
I just made this repo public where I have my [nvim-dotfiles](https://github.com/fdncred/nvim-dotfiles). I confess. I'm not good with nvim so I'm pretty sure I've messed something up.

mrdgo:
aargh, your setup looks fine - I can't reproduce! Revert is probably the best option for now.

fdncred:
ok, i feel bad to revert but it's probably for the best right now until we can figure out a way forward.

mrdgo:
No it's fine. I'll also try to add integration tests w/ nvim - maybe we can avoid issues like this in the future

PR #114, opened 2024-09-24
feat: raw strings: distinguish delimiters from actual string

mrdgo:
We could set the delimiters to a different highlight group. Example `@punctuation.special`:
image

To see all available groups, hit `:h treesitter-highlight-groups`. I could see:
- `@punctuation.special`
- `@punctuation.delimiter`
- `@tag.delimiter`
- `@string.escape`
- `@string.special`
- `@string.special.symbol`

Any opinions?

Edit: with highlights now in `nvim-treesitter`, this feels a bit weird. But we can still maintain our own highlights :D


fdncred:
Thanks

PR #117, opened 2024-09-25
feat: reintroduce raw-strings

mrdgo:
[Discussion with tree-sitter devs](https://github.com/tree-sitter/tree-sitter/discussions/3878#discussioncomment-11139978)

This will break the current `nvim-treesitter` config. They have to update to also use `scanner.c`. People using that config might want to use "our" way to install ts-nu.

After merge, it might be necessary to run `:TSUpdate nu` at least once. I don't know about treesitter's caching, so let's just hope that nobody's setup breaks.

With that scanner, we can also create custom logic for unquoted strings.


mrdgo:
@blindFS
Feel free to try this yourself! The code stands, it passes tests and correctly highlights in neovim-nightly. Please follow the link in the description (discussion with tree-sitter-devs) and see the current show-stopper. As soon as I start the exact same setup within neovim 0.10.2/0.10.1/0.9.5, it crashes without any error message. The only thing that I found out is a null-pointer dereference - which doesn't make sense, because the scanner works standalone and in neovim-nightly. I am deeply puzzled and don't really have an idea how to procceed

siph:
fwiw this is working for me (`3c50ee4186153dd768141c9d0a2fcc1fdcd47b03`) in `v0.10.2`:
Screenshot_20241120_144941

Screenshot_20241120_145028

I'm using nix to build the grammar and neovim.

blindFS:
I also have no problem with your changes made on-top of current main branch:

image

image

My neovim is installed from nix-darwin for macos.

tree-sitter.json
mrdgo, line 1: Is this file even required? Just noticed that I pushed it.

src/scanner.c
```diff
@@ -0,0 +1,172 @@
...
+) {
+    if (valid_symbols[ERROR_SENTINEL]) {
+        return false;
+    }
+
+    Scanner *s = (Scanner *) payload;
+    lexer->log(lexer, "Nu Scanner: level [%i], emitted [%b]\n", s->level,
+               s->emitted_content);
+
+    if (valid_symbols[RAW_STRING_BEGIN] && s->level == 0) {
```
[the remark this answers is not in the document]
mrdgo: done

tree-sitter.json
clason: It's newly created by the latest tree-sitter CLI. (Please make sure you always use a released version of the CLI -- 0.24.4 right now -- not a prerelease commit since downstream consumers will not have adapted to the breaking changes that will come with tree-sitter 0.25.)

.github/neovim/init.lua
```diff
@@ -0,0 +1,37 @@
+-- Bootstrap lazy.nvim
+local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
```
clason, line 2: You don't need `lazy`; just clone and set up nvim-treesitter directly.

You might also want to consider using the `main` branch for tests; that is simpler to set up and has much better performance (but still WIP), see https://github.com/tree-sitter/tree-sitter/blob/master/.github/workflows/nvim_ts.yml

Make sure you include a `generate` step!

(You can call `./scripts/install-parsers.lua nu` to only install your parser.)

Check out https://github.com/tree-sitter/workflows as well.

.github/workflows/main.yml
```diff
@@ -30,3 +30,9 @@ jobs:
 
       - name: test
         run: npx tree-sitter test
+
+  tests:
+    uses: ./.github/workflows/nvim-tests.yml
+    with:
+      neovim_versions: |
+        [ "nightly", "v0.10.2", "v0.10.1", "v0.9.5" ]
```
clason, line 38: nvim-treesitter no longer supports v0.9 and earlier.

mrdgo: thanks!

.github/neovim/init.lua
mrdgo: I am not happy with the pipeline, you reinforced this gut feeling. I will take a look, thanks!

src/scanner.c
```diff
@@ -0,0 +1,172 @@
...
+            return true;
+        } else if (level == s->level && len == 0) {
+            return false;
+        }
+    }
+
+    return false;
+}
+
+bool scan_raw_string_end(TSLexer *lexer, Scanner *s) {
```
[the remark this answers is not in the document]
blindFS: I can pass all your test cases with the following code:
```c
static bool scan_raw_string_end(TSLexer *lexer, Scanner *s) {
    lexer->log(lexer, "END\n");
    adv;
    for (unsigned i = 0; i < s->level; i++) {
        adv;
    }
    s->level = 0;
    return true;
}
```

image

Better to put the `raw-strings.nu` in `expr` ?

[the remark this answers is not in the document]
blindFS: Let me try to explain my understanding of the logic here:

1. in `scan_raw_string_content` we already make sure the ending level is correct,
   - if it's not, for example, at the 2nd  `'` in `r##' hello'# '##`, it won't stop the while loop, it will advance further until the last `'` is reached, then mark_end there.
2. then in `scan_raw_string_end`, we simply advance `s->level` + 1 (for the `'`) times.

clason:
And in general, I would strongly recommend to make conceptually unrelated changes in separate PRs; that makes it much easier to review and merge preliminary steps.

(You can cherry-pick commits from PRs that are required to be merged first, as long as you clearly label them as such.)

mrdgo:
Incorporated all reviews, removed the pipeline changes that neither provide insights nor prevent anything. Maybe pipeline is a separate issue that should be addressed at some point.

fdncred:
Ping me if/when we're ready to land this and I'll push the button.

mrdgo:
Thanks to @blindFS for the thorough review! Is there someone else who would want to review? Otherwise, I'd be ready to land, again

mrdgo:
Aaah, wait a second

src/scanner.c
```diff
@@ -0,0 +1,163 @@
...
+        lexer->log(lexer, "Set mark\n");
+        lexer->mark_end(lexer);
+        len += num;
+        adv;
+        uint8_t level = consume_chars(lexer, '#');
+        lexer->log(lexer, "Consumed [%i] #\n", level);
+        if (level == s->level) {
+            lexer->log(lexer, "Detected end\n" );
+            return true;
+        } else if (level == s->level && len == 0) {
```
blindFS, line 115: No need for this branch now.

src/scanner.c
```diff
@@ -0,0 +1,163 @@
...
+        skip;
+    }
+}
+
+static bool scan_raw_string_begin(TSLexer *lexer, Scanner *s) {
+    lexer->log(lexer, "BEGIN\n");
+    // scan for r#' r##' or more #
+    skip_whitespace(lexer);
+
+    if (lexer->lookahead == 'r') {
```
blindFS, line 86: This check is already done.

src/scanner.c
mrdgo: true, removed

src/scanner.c
mrdgo: This also makes `skip_whitespace` obsolete, thanks!

queries/nu/highlights.scm
```diff
@@ -90,6 +90,10 @@ file_path: (val_string) @variable.parameter
 (escaped_interpolated_content) @string
 (expr_interpolated ["(" ")"] @variable.parameter)
 
+(raw_string_begin) @string
```
blindFS, line 93: We probably don't need this, `val_string` will get it properly highlighted.

mrdgo: True, removed

fdncred:
looks like we're closer, just have a conflict now.

fdncred:
Are we ready?

blindFS:
> Are we ready?

I think so.

mrdgo:
@clason does nvim-ts' lockfile interfere, even when I override the `install_info`?

clason:
Yes, the lockfile takes precedence. (Nvim-ts is not a general purpose installer; you can shoehorn additional parsers but it's not designed for replacing tracked parsers.)

mrdgo:
> Yes, the lockfile takes precedence. (Nvim-ts is not a general purpose installer; you can shoehorn additional parsers but it's not designed for replacing tracked parsers.)

Would you then pretty-please be so kind :innocent:

clason:
We have automation for that

PR #120, opened 2024-09-26
fix: highlights artifact from flags #110

mrdgo:
In #110, I changed highlights. I accidentally committed `(val_string) @variable.parameter`, which should actually be `@string`.


fdncred:
Thanks!

Issue #128, opened 2024-10-31
More accurate command arguments?


blindFS:
Background: The official formatter nufmt seems broken to me, and I'm trying to implement another nu script formatter based on [topiary](https://github.com/tweag/topiary) and this tree-sitter-nu grammar.

Then I noticed there're many parsing conflicts in command arguments for example:

```nu
$a | get 0.username # 0.us parsed as val_duration and 'ername' as string, same thing for val_bool/val_filesize etc.
echo foo('bar')baz # parsed down to command with 3 arguments (should be 1)
```

It seems to me that as a whitespace sensitive shell language, external scanner is a mandatory to make it right according to https://github.com/tree-sitter/tree-sitter/discussions/2159

If you have any better idea of how to fix these or you owners are conservative about this kind of changes, I will on-hold my effort of writing the external scanner.

fdncred:
The `nufmt` formatter is definitely incomplete and won't work in (probably) a lot of situations. I also looked at using tree-sitter in topiary as a formatter and it just seems pretty complex the more I looked at it.

I'm up for whatever makes it work and is supportable.

blindFS:
@fdncred I got a [prototype](https://github.com/blindFS/topiary-nushell) of topiary formatter here. The biggest issue for now, in my view, is these command argument corner cases, since formatting will break semantics if not parsed accurately. Hopefully an external scanner can solve all of them at once.

fdncred:
oooh, I'll definitely take a look at that. Really appreciate your help with this!

PR #132, opened 2024-11-06
More accurate unquoted command args.

blindFS:
It turns out that external scanner is not necessary to fix most cases in #128 .

One existing test case modified as the original entry is actually parsed as string in nu 0.99.1:

image

More corner cases added to test.


fdncred:
Thanks for working on this!

PR #133, opened 2024-11-06
Special treatment for filesize `0b`

blindFS:
This PR is a partial fix for file size `0b` which will lead to parse error in current grammar.

image

Since `0b` is a special single terminal token in `$.val_number` and `$.val_binary`.
A perfect split into `value` 0 and `unit` b needs unnecessarily complicated tweaking of potential conflicts (also need to split `0b` in other 2 rules).

 With this PR however, `0b` is added to `$.val_filesize` as a special case, no detailed fields generated:

```scheme
  (pipeline [0, 0] - [1, 0]
    (pipe_element [0, 0] - [0, 2]
      (val_filesize [0, 0] - [0, 2])))
```

But it avoids the error with minimal effort.


fdncred:
ya, that's probably a good strategy. thanks!

PR #134, opened 2024-11-07
Fix unquoted command args with leading +

### LINKED ISSUE #131: Unquoted strings starting with + not recognized as strings
The most obvious example I can think of:

`^chmod +x $file`

It's not a huge problem, but it does show up as an error. Interestingly, if I change it to this, it recognizes it as a string:

`^chmod x+ $file`

I looked into it a little bit in `_unquoted_rule` in `grammar.js`, but I'm a bit scared of unintended consequences by modifying that part, so I thought I'd at least create this issue in case someone who has worked on this can see a simple fix.

blindFS:
This PR fixes #131  by grouping "+/-" and digits in decimals.
Related test cases in `cmd-026-unquoted-string-with-leading-plus`.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1496,8 +1496,9 @@ function _decimal_rule(immediate) {
     choice(
       seq(head_token(/[\d_]*\d[\d_]*/), optional(exponent)),
       seq(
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks!

PR #135, opened 2024-11-07
Fix assignment with pipeline as rhs

### LINKED ISSUE #121: Fail to parse assignments with pipeline
Code snippet like this
```nu
$a = $a
  | str substring ..3
```
leads to parsing error because the right hand side is a pipeline instead of expression.
This is involved by #28 

```nu
=====
assignment-003-assignment-to-a-pipeline
=====

$x += 1 | $in + 10
# Note that rhs of ++= is not a pipeline, but only 1
# currently nushell parses this as two statements 
# $x ++= 1 and  $in + 10. Therefore you will get error
# about adding int to nothing, and $x increased by one

-----

(nu_script
  (assignment
    (val_variable
      (identifier))
        (val_number))
    (ERROR)
      (pipeline
        (pipe_element
          (expr_binary
            (val_variable)
            (val_number))))
    (comment)
    (comment)
    (comment)
    (comment))
```
Since this is no longer the truth, current version of nushell takes `1 | $in + 10` as the rhs value, should we consider to change it back to $.pipeline ?

blindFS:
This PR fixes #121 
Related test cases reverted to pre #28 state, which is more aligned with current nushell language.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -683,28 +683,6 @@ module.exports = grammar({
         BRACK().close_brack,
       ),
 
-    /// Assignment Statement
-
-    assignment: ($) => {
-      const opr = [
-        PUNC().eq,
-        OPR().assign_add,
-        OPR().assign_sub,
-        OPR().assign_mul,
-        OPR().assign_div,
-        OPR().assign_append,
-      ];
-
-      return prec.left(
-        PREC().assignment,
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks for cleaning this up. We really appreciate all your effort here.

PR #136, opened 2024-11-08
Allow unquoted values in val_record

blindFS:
This PR fixes situations where record value is an unquoted string

image

Also cleaned list_entry rule a little bit since signed numbers, flags and unquoted string start with `+/-` are more easy to distinguish.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -437,12 +437,7 @@ module.exports = grammar({
     match_guard: ($) => seq(KEYWORD().if, $._expression),
 
     _match_pattern_expression: ($) =>
-      choice(
-        $._match_pattern_value,
-        alias($._expr_unary_minus, $.expr_unary),
-        $.val_range,
-        $.expr_parenthesized,
-      ),
+      choice($._match_pattern_value, $.val_range, $.expr_parenthesized),
 
     _match_pattern_value: ($) =>
       choice(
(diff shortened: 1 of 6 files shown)


blindFS:
BTW, I don't know whether this is desired:

image

But it makes unquoted strings easier to parse and current grammar works in the same way.
If so, `_expr_unary_minus` doesn't make much sense then.

fdncred:
Is you screenshot the "before" or "after" this PR?

x is supposed to be a string here
```nushell
❯ {k: a}
╭───┬───╮
│ k │ a │
╰───┴───╯
```
```nushell
❯ {k: a} | describe -d | table -e
╭─────────┬────────────────╮
│ type    │ record         │
│         │ ╭───┬────────╮ │
│ columns │ │ k │ string │ │
│         │ ╰───┴────────╯ │
╰─────────┴────────────────╯
```

blindFS:
> Is you screenshot the "before" or "after" this PR?
> 
> 
> 
> x is supposed to be a string here
> 
> ```nushell
> 
> ❯ {k: a}
> 
> ╭───┬───╮
> 
> │ k │ a │
> 
> ╰───┴───╯
> 
> ```
> 
> ```nushell
> 
> ❯ {k: a} | describe -d | table -e
> 
> ╭─────────┬────────────────╮
> 
> │ type    │ record         │
> 
> │         │ ╭───┬────────╮ │
> 
> │ columns │ │ k │ string │ │
> 
> │         │ ╰───┴────────╯ │
> 
> ╰─────────┴────────────────╯
> 
> ```

The screenshot with error is before this PR, should be fine after this.

blindFS:
`_expr_unary_minus` however is left as it was, which is not exactly current nushell's behaviour

fdncred:
Thanks

PR #137, opened 2024-11-09
Fixed some minor errors

blindFS:
This PR fixed following issues in current grammar:
1. hide-env with flags

image

2. do + closure + more than one arguments / parenthesized multiline do statement

image

3. wrong result for multiline `expr_interpolated` (here `bar` should be argument instead of the command head of another pipeline)

image

4. hide those `"\n"` terminal tokens, which makes it easier for query, also more windows friendly.


fdncred:
wow, thanks!

PR #139, opened 2024-11-12
Manually control all newline characters

### LINKED ISSUE #122: Fail to parse multi-line binary expression
Failed case:

```nu
('foo'

+

'bar')
```

I can manage to make it work by adding additional `expr_binary_parenthesized` rule like this:

```diff
diff --git a/grammar.js b/grammar.js
index 5ad1222..08a5d00 100644
--- a/grammar.js
+++ b/grammar.js
@@ -26,6 +26,7 @@ module.exports = grammar({
     [$._val_number_decimal],
     [$._immediate_decimal],
     [$.expr_parenthesized],
+    [$.expr_binary_parenthesized],
     [$.val_variable],
     [$._expression, $._expr_binary_expression],
     [$._match_pattern_value, $._value],
@@ -577,6 +578,7 @@ module.exports = grammar({
       seq(
         choice(
           prec.right(69, $._expression),
+          alias($.expr_binary_parenthesized, $.expr_binary),
           $._ctrl_expression_parenthesized,
           $.where_command,
           alias($._command_parenthesized_body, $.command),
@@ -592,6 +594,7 @@ module.exports = grammar({
     pipe_element_parenthesized_last: ($) =>
       choice(
         $._expression,
+        alias($.expr_binary_parenthesized, $.expr_binary),
         $._ctrl_expression_parenthesized,
         $.where_command,
         alias($._command_parenthesized_body, $.command),
@@ -818,6 +821,30 @@ module.exports = grammar({
         ),
       ),
 
+    expr_binary_parenthesized: ($) =>
+      choice(
+        ...TABLE().map(([precedence, opr]) =>
+          prec.right(
+            // @ts-ignore
+            precedence,
+            seq(
+              field("lhs", $._expr_binary_expression),
+              "\n",
+              // @ts-ignore
+              field("opr", opr),
+              field(
+                "rhs",
+                choice(
+                  $._expr_binary_expression,
+                  alias($.expr_binary_parenthesized, $.expr_binary),
+                  alias($.unquoted, $.val_string),
+                ),
+              ),
+            ),
+          ),
+        ),
+      ),
+
     _expr_binary_expression: ($) =>
       choice($._value, $.expr_binary, $.expr_unary, $.expr_parenthesized),
```

But there are probably better ways to do it.

blindFS:
While attempting to fix #122 , I found the ambiguity of newline characters is a big obstacle for multiline expressions to be parsed accurately.

Current grammar put newline characters in `extras` (`\s`), allowing arbitrary numbers of them to be present freely between any nodes. While some rules specifically announced newline characters, making them selected with higher priorities when a newline is matched in the exact sequence.

This makes adding the rule for `expr_binary_parenthesized` extremely difficult since newline characters breaks the precedence order.

This PR did quite a lot changes:
1. remove newline from extras
2. manually add newline to related rules
3. several existing test cases are broken, but the new results seem to be more accurate, so I changed the target results
  - most of them is about the position of comments
  - `val_record` as register signature
4. more multiline related test cases added

⚠️ There are probably new errors introduced by this PR!
Please take time to review it with caution.
I'll continue to test it on more nushell scripts to make it more robust.

If you guys have better solutions, I'm more than happy to switch.


test/corpus/stmt/register.nu
```diff
@@ -54,74 +54,59 @@ register ~/plugins/nu-plugin-math {
...
-
-register $plugin
-
------
-
-(nu_script
-  (stmt_register
-    (val_variable
-      (identifier))))
+    (val_string)
```
blindFS, line 57:
```js
    stmt_register: ($) =>
      prec.left(
        -1,
        seq(
          KEYWORD().register,
          field(
            "plugin",
            choice(alias($.unquoted, $.val_string), $.val_variable),
          ),
          field("signature", optional($.val_record)),
        ),
      ),
```

The new result is more accurate for the rule

fdncred:
I really don't have a problem moving forward with this because even if it breaks something, you've been super helpful to fix those type of things. If this is going to make things better in the long run, a little temporary instability is a small price to pay.

blindFS:
@fdncred I think this is helpful to decouple intertwined rules and enable finer grained control of the grammar, on the other side, it forces us to write more intricate rules when newline characters are allowed in them.

Let's wait to see whether other people @CabalCrow @mrdgo have different opinions on this, during the on-going test on a broader range of scripts.

CabalCrow:
Back when I was attempting a rewrite, I was trying to put the newlines in scanner due to the same issue. There were plenty of problem with comments and all other types of syntax that is allowed by nushell.

Do note that the bash tree sitter also had some of the issue, so maybe this could be a limitation of what you can do with tree-sitter without relying on the scanner.

blindFS:
@CabalCrow Yeah, external scanner was my initial thought and probably is the ultimate solution. Yet it turns out that most of known errors can still be solved without a scanner, comments are definitely on the list of attention here. I feel that keeping it all js for now helps to recognize the parts that really requires a scanner, for example the unquoted string rule already looks insanely complicated.

mrdgo:
Thanks for the PR and the request for review.

I honestly feel some pain looking at all the places where you now explicitly allow newlines. On the other hand, it's a great advantage to solve it in js. I'll take a closer look tomorrow.

test/corpus/decl/use.nu
```diff
@@ -52,7 +52,28 @@ use $"('s' + 't' + 'd')"
         (pipeline
           (pipe_element
             (expr_binary
-              (val_string)
```
blindFS, line 55: nested binary expression made left associative.

mrdgo (approved):
Thanks for all the cleanup work alongside the newline control :)

grammar.js
```diff
@@ -5,63 +5,64 @@ module.exports = grammar({
...
+    [$._terminator],
+    [$._val_number_decimal],
+    [$.block, $.val_closure],
+    [$.block, $.val_record],
+    [$.ctrl_if_parenthesized],
+    [$.ctrl_try_parenthesized],
+    [$.expr_binary_parenthesized],
+    [$.nu_script, $._terminator],
+    [$.pipeline],
+    [$.pipeline_parenthesized],
```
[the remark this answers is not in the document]
blindFS: For single-item lists, Treesitter will figure out how many times should each pattern repeat.

grammar.js
```diff
@@ -160,32 +166,53 @@ module.exports = grammar({
...
-          field("param_long_flag", $.param_long_flag),
-          field(
-            "param_long_flag",
-            seq($.param_long_flag, field("flag_capsule", $.flag_capsule)),
-          ),
-        ),
-        optional(
+      prec.right(
+        seq(
+          repeat($._newline),
```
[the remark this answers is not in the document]
blindFS: Since `$.parameter` is repeated, I feel it more clear that it only allows edge newlines on one side (here on the beginning). Shared newline across parameters probably wont cause any conflict though.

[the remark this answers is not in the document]
mrdgo: I hope that tests wouldn't pass on conflict, I was just wondering. Consider it more a curious question than a request for change :)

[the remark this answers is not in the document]
blindFS: Actually, it's a good point, I probably need to use a helper function for those kind of repeated patterns, somewhat like list_entries.

grammar.js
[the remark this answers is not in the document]
mrdgo: I read the docs and didn't understand, thanks for your explanation - now I got it :)

blindFS:
I think it is ready for review.
I have tested it on all nu scripts in [nushell/nu_scripts](https://github.com/nushell/nu_scripts)

Either it is parsed with no error (probably wrong answer though)
or the error is rooted somewhere else (collected a dozen of new issues :-( ).

I left some parenthesized rules unchanged by intention, technically newlines should be allowed in them, but I think nobody writes nushell in a crazy way like this:

```nushell
(
let
foo
=
1
)
```

The CI action pipeline has permission issues at the time.

grammar.js
```diff
@@ -92,7 +92,8 @@ module.exports = grammar({
     // manually controlled by adding the following to parenthesized rules
     _newline: (_$) => /\r?\n/,
     _terminator: (_$) => choice(PUNC().semicolon, _$._newline),
-    _separator: (_$) => choice(/[ \t]+/, _$._newline),
+    _space: (_$) => /[ \t]+/,
+    _separator: (_$) => choice(_$._space, _$._newline),
```
mrdgo, line 96: Woudn't `_$` indicate that the parameter is unused? `_terminator` seems to violate this assumption as well. What is your reasoning behind it?

blindFS: I thought the naming of `_$` means terminal token

blindFS: @mrdgo you're right, referencing to other named rules makes them non-terminal then. I'll change the naming

mrdgo: Well, then we're both right, aren't we? :D

blindFS: Yeah, good catch

blindFS:
Github action file modified to git rid of permission error in the last commit. @fdncred

mrdgo:
I am absolutely fine with this!

As soon as we have a `scanner.c`, someone could try to implement newline support in there and compare. But I don't really see any show-stopper, here.

fdncred:
Thanks. Let's move forward and see how it goes. Appreciate all the work here.

Issue #142, opened 2024-11-16
Errors detected from nushell/nu_scripts


blindFS:
## Command

* [ ] - executable path as the command head
  - example: `~/test.nu arg1 arg2`
* [ ] - val_interpolated as the external command head
  - example: `^$"($cmd)"`

## Identifier

* [ ] - command identifier too restricted, example:
  - `g++ foo.cpp`
  - `7z`
  - `export module foo/`
* [ ] - variable identifier, example:
  - `$0`
* [ ] - record key should allow more punctuations
  - example: `{%: foo}`

## Redirection

* [ ] - append ops, example `1 o>> foo`
* [ ] - values redirection, example `1 + 1 o> foo`
* [ ] - redirect pipes should be single token, example `cmd1 e>| cmd2`

## Parameters/command args

* [ ] - single `-` should be allowed as short_flag
* [ ] - short_flag_equals_value: `-flag=foo`
* [ ] - unquoted as default value in parameters: `--flag=name:string@"nu-complete flag"`
* [ ] - list types with completion info: `--flag(-f): list<string@"nu-complete flag">`
* [ ] - spread types: #98 

## Misc

* [ ] - variable as cell_path: `$foo.$bar.$baz`
* [ ] - unquoted strings with more punctuations: `echo foo"bar"`

fdncred:
if anyone wants to work on any one of these specific issues, let me know and i can put your name by the item.

PR #143, opened 2024-11-17
Fix spread operators & redirections

### LINKED ISSUE #98: Support the spread operator
image

Currently, the `...` is parsed as a string, whereas it's a special operator that should be assigned something special, and then highlighted as an operator. If it's parsed as a `expr_unary`, `highlights.scm` can then be changed as follows:

```diff
diff --git a/queries/nu/highlights.scm b/queries/nu/highlights.scm
index c73f8f3..ce0b839 100644
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -161,7 +161,7 @@
     "++="
 ] @operator)
 
-(expr_unary ["not" "-"] @operator)
+(expr_unary ["not" "-" "..."] @operator)
 
 (val_range [
     ".."
```

blindFS:
This PR should fix #98, as well as several easy problems mentioned in #142.

I think I should keep each PR focused for better traceability, so the trickier issues related to identifier and parameters are left for later.

## Command

* [x] - val_interpolated as the external command head
  - example: `^$"($cmd)"`

## Redirection

* [x] - append ops, example `1 o>> foo`
* [x] - values redirection, example `1 + 1 o> foo`
* [x] - redirect pipes should be single token, example `cmd1 e>| cmd2`

## Parameters/command args

* [x] - spread types

## Misc

* [x] - variable as cell_path: `$foo.$bar.$baz`, it turns out `$bar` is treated as string instead.

## Changes made to the existing test cases

- `unquoted` node in `decl_use` renamed to `val_string` to align with `stmt_source`
- cases that should fail now simply tagged with `:error` since we generally don't care where the error occurs

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -16,7 +16,7 @@ module.exports = grammar({
     [$._expression, $._expr_binary_expression],
     [$._expression_parenthesized, $._expr_binary_expression_parenthesized],
     [$._immediate_decimal],
-    [$._match_pattern_expression, $._list_item_expression],
+    [$._match_pattern_expression, $._item_expression],
     [$._match_pattern_list, $.val_list],
     [$._match_pattern_record, $.val_record],
     [$._match_pattern_record_variable, $._value],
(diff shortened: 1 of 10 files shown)


fdncred:
Nice work, thanks!

PR #147, opened 2024-11-19
Fix several issues of identifiers

blindFS:
This PR fixes most of the remaining issues related to identifiers as reported in #142 .

```nushell
# values
null
7b
nan
# commands
g++
7z
7mss
truee
```

It's still not perfect, nushell handles special characters like `-, #, ., ()` in identifiers very dynamically.
I think this is the time to finally resort to an external scanner for further accuracy.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -9,6 +9,7 @@ module.exports = grammar({
 
   inline: ($) => [
     $._do_expression,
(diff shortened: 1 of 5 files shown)


fdncred:
Thanks

PR #149, opened 2024-11-20
Fix single-use env and other fixes

### LINKED ISSUE #146: setting env vars bash style produces errors
This is from nushell's toolkit.nu file
image

You can see that setting an env var and running a command, bash style, produces an error.
```nushell
_EXTRA_=bin nu .github/workflows/release-pkg.nu
```

blindFS:
This PR
1. fix #146 , check the test cases in the new env.nu file, remind me if any format is missing.
2. force whitespace in between of binary expression: `1+1` should be command
3. simplify the unquoted rule with lexical precedence.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -24,9 +24,7 @@ module.exports = grammar({
 
   conflicts: ($) => [
     [$._binary_predicate_parenthesized],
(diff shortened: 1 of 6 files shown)


fdncred:
Thanks for working on this!

PR #150, opened 2024-11-21
Fix ctrl_match with _entry_separator

blindFS:
This PR fixes some errors of `ctrl_match` because of non-explicit separator in match body.

image


src/scanner.c
mrdgo, line 1: hmmm

fdncred: do you mean to have scanner.s included in this PR?

blindFS: It's an accident, forgot to remove it from the branch...

fdncred:
Thanks

PR #153, opened 2024-11-22
Precedence tuning

blindFS:
This PR

1. fixes several corner cases, such as:
```nushell
# empty list
[,,,]

# record with leading comment
# previously parsed as closure
{
# foo
echo : 1
}
```
2. generalizes `general_body_rule` to cover more rules within the same pattern
3. makes anonymous tokens more friendly for topiary to format


fdncred:
Thanks

PR #154, opened 2024-11-24
Reverted: force separator around binary operators

blindFS:
This PR reverts this trick introduced by #149 :

```js
/**
 * group operator and its preceding/succeeding whitespace/newline
 * @param {string} opr
 * @param {boolean} parenthesized
 */
function operator_with_separator(opr, parenthesized) {
  const sep = parenthesized ? choice(/[ \t]/, /\r?\n/) : /[ \t]/;
  return alias(token(seq(repeat1(sep), opr, repeat1(sep))), opr);
}
```

It turns out that this will cause unwanted side effects of blurring boundaries of tokens around `binary_expression`:

image

compared to the ideal result after this PR:

image

Instead, `cmd_identifier` is tweaked to keep expressions like `1+1` away from being parsed as `binary_expression`.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -30,6 +30,7 @@ module.exports = grammar({
     [$._block_body, $.record_body, $.val_closure],
     [$._block_body, $.shebang],
     [$._block_body, $.val_closure],
+    [$._expression, $._expr_binary_expression],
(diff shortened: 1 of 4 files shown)


fdncred:
good catch. thanks!

Issue #119, opened 2024-09-26
add textobjects.scm


mrdgo:
With a `textobjects.scm`, we can support additional tree-sitter-based motions, such as `@function.inner`. More motivation incoming.

[nvim-treesitter-textobjects](https://github.com/nvim-treesitter/nvim-treesitter-textobjects)

mrdgo:
Closing this because of [this PR](https://github.com/nushell/tree-sitter-nu/pull/159).

If you stumble across this and want to request a new query, feel free to open a new issue and we can look into it.

PR #159, opened 2024-11-28
feat: add textobjects.scm

mrdgo:
This PR implements the first queries for #119.

With this sample config, you can hit `yaf` (yank around function) to copy the closest function, `dif` (delete in function) to clear the contet of the closest function, with sourround-like plugins `ysaSb` (sourround around  "()"-block) to sourround a statement with parentheses, or whatever keys you map to what query.

```lua
textobjects = {
    select = {
        enable = true,
        keymaps = {
            -- You can use the capture groups defined in textobjects.scm
            -- For example:
            -- Nushell only
            ["aP"] = "@pipeline.outer",
            ["iP"] = "@pipeline.inner", -- selects the closest pipeline element

            -- supported in other languages as well
            ["af"] = "@function.outer",
            ["if"] = "@function.inner",
            ["al"] = "@loop.outer",
            ["il"] = "@loop.inner",
            ["aC"] = "@conditional.outer",
            ["iC"] = "@conditional.inner",
            ["iS"] = "@statement.inner",
            ["aS"] = "@statement.outer",
        }, -- keymaps
    }, -- select
}, -- textobjects
```

To me, these queries are sufficient. I wouldn't setup any other ones, so I'd declare this as ready to merge. Feel free to request additional queries.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ on:
       - main
 jobs:
   main:
-    name: install
+    name: setup, lint, test
 
     runs-on: ${{ matrix.os }}
     strategy:
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let package = Package(
                 ],
                 sources: [
                     "src/parser.c",
-                    // NOTE: if your language has an external scanner, add it here.
+                    "src/scanner.c",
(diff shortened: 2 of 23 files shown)


fdncred:
Looks fun! Thanks!

PR #160, opened 2024-11-28
Feat: support commands with spaces

mrdgo:
- **chore: migrate to new tree-sitter.json**
- **experiment: place mean regex**


blindFS:
How about [this way](https://github.com/nushell/tree-sitter-nu/pull/156#issuecomment-2506453732) ?

Changing the identifier is dangerous since it's used in other places.

fdncred:
Looks the same to me, highlights and syntax tree. But maybe I'm doing something wrong. I'm not sure how to tell if I'm using the latest changes. My nvim configuration for tree-sitter-nu points to my local tree-sitter-nu repo.
image

blindFS:
Sorry, I wasn't aware that the whole idea of long regex here is to sync with those generated for vscode.

But I think we have easier way to generate our own copy, such as 
`scope commands | get name | where {' ' in $in} o> list`

And then we can read it from grammar.js to do whatever we want.

mrdgo:
1. this thing is not working yet, just a heads up that I pursue the regex idea
2. yes, @blindFS this is about reusing the regex that is automatically created for vscode
3. If we decide to do this, I'd love to build some automation to automatically sync the regex from the other repo. But one thing after the other.

mrdgo:
> Looks the same to me, highlights and syntax tree. But maybe I'm doing something wrong. I'm not sure how to tell if I'm using the latest changes. My nvim configuration for tree-sitter-nu points to my local tree-sitter-nu repo. image

yes, there is another problem regarding parsers: you have to fork nvim-treesitter and set nu's config to your local checkout. Maybe even update the lockfile to the currently checked out commit, but I am not 100% sure about that.

PR #161, opened 2024-11-29
Feat: internal commands with spaces

blindFS:
For test purpose #156


fdncred:
Seems to work pretty good to me
image

blindFS:
@fdncred yes, it works, but the dirty hack somehow worries me a little bit :-)

fdncred:
I just noticed that the INTERNAL only has commands with spaces in them vs the entire list of commands. Is that because the single word commands are already captured elsewhere?

blindFS:
> I just noticed that the `INTERNAL` function only has commands with spaces in them vs the entire list of commands. Is that because the single word commands are already captured elsewhere?

Yes, as the `cmd_identifier`.

fdncred:
Also keep in mind that in order to get commands and plugins with spaces in them, you have to have the plugins built, installed, and registered for this command below to work. Also, that command would get multiword aliases and custom commands too.
```nushell
scope commands | get name | where {' ' in $in} o> list
```
This is the command below is probably more appropriate. It could be easily tweaked to include only the commands/plugin commands with spaces in them. We could also include `command_type == external` and that adds whatever custom completions you have loaded. 🤔 
```nushell
# generated with Nu 0.100.1
help commands | 
   where command_type == built-in or command_type == plugin or command_type == keyword |
   get name | 
   where {' ' in $in}
```

blindFS:
Current implementation can't tell internals/externals from each other, if we want them highlighted differently, more complicated `cmd_identifier` is necessary.

fdncred:
I'm not saying that they should be highlighted differently. I'm just saying that the help commands or scope commands only needs to list the appropriate things. e.g. custom commands should not be included in our list.

blindFS:
> I'm not saying that they should be highlighted differently. I'm just saying that the help commands or scope commands only needs to list the appropriate things. e.g. custom commands should not be included in our list.

Sure, this pr is just a demo, we should redesign pipeline to generate the list.

fdncred:
I have the `rustup`, `cargo`, `git` and `gh` custom completions loaded (along with keywords, builtins and plugins) and I have a list of 531 commands with spaces in them. I can provide that list or just update this PR before landing it.

blindFS:
Does your script above generate all 531 of those commands? If so, we can add that script along with the generated list file to the repo. But I guess only you can maintain that list right? We probably only get a small part of it with the same script?

blindFS:
We probably should test the impact on performance, 531 seems to be a significant number.

blindFS:
@fdncred Do we need `cargo xxx` in the list?

fdncred:
We can remove any of them that we want. However, I'm not sure 19 commands will make any difference. These are just the custom completions I usually run with `rustup`, `cargo`, `git` and `gh`.

blindFS:
@fdncred we can keep it and update the test case,

the performance loss is about 7~8%, `4757/5136 bytes/ms`

PR #163, opened 2024-11-29
Fix: test case update

blindFS:
I was too slow on the thread of #161 , some necessary updates is not pushed on-time.

1. commands start with `overlay` and `export` are removed, they are treated as keywords
2. one test case with `cargo install` updated


blindFS:
@fdncred command list moved outside of grammar.js into `internal.txt`.

steps to update the list:

1. run your script (remember to exclude those starts-with 'export' or 'overlay')
2. save the result to `internal.txt`
3. run `tree-sitter generate` to update the grammar

fdncred:
> I was too slow on

LOL. Sorry, like to pull the trigger fast.

Are overlay and export still identified and highlighted correctly when they have spaces in the command names?

blindFS:
> > I was too slow on
> 
> LOL. Sorry, like to pull the trigger fast.
> 
> Are overlay and export still identified and highlighted correctly when they have spaces in the command names?

image

I think so, there're rules specified for them in highlights.scm.

fdncred:
Just to document my command somewhere, here it is.
```nushell
help commands |
   where command_type == built-in or command_type == plugin or command_type == keyword or command_type == external |
   where name not-like overlay and name not-like export | get name |
   where {' ' in $in} | each {|r| $'"($r)"'} | save -f internal.txt
```
I really don't like having multiple where's but it's easier to read this way

blindFS:
> Just to document my command somewhere, here it is.
> 
> ```nushell
> help commands |
>    where command_type == built-in or command_type == plugin or command_type == keyword or command_type == external |
>    where name not-like overlay and name not-like export | get name |
>    where {' ' in $in} | each {|r| $'"($r)"'} | save -f internal.txt
> ```
> 
> I really don't like having multiple where's but it's easier to read this way

`each {|r| $'"($r)"'}` this should be removed.

PR #166, opened 2024-12-06
fix: multiline collection type and more

blindFS:
This PR fixes:

1. `param_value` before `param_type`, e.g. `--param=value: string`
2. `error make {}` by removing it from internal.txt
3. `val_range` in binary expression, e.g. `1 in 1..3`
4. multiline type signature, e.g.

```nushell
table<
foo: string
>
```


fdncred:
looks good. thanks!

PR #167, opened 2024-12-10
fix: whitespace issues of if/for/error/alias

blindFS:
This PR fixes some easy ones of the whitespace issues:

image

image

There're other harder cases remaining.

I don't quite get the logic behind, seems like bugs of `tree-sitter` itself to me. The work-around solutions in this PR were found through bisection and try-and-error.


fdncred:
Thanks

PR #171, opened 2024-12-17
fix: performance issue of comment doc queries

### LINKED ISSUE #170: 100% CPU usage in new example configuration
(to be clear, 100% of one CPU core; but that still feels incredibly laggy)

image

starting at this line https://github.com/nushell/nushell/blob/39770d41975a0dae3cdc89802215416a4c6671e4/crates/nu-utils/src/default_files/sample_config.nu#L376

blindFS:
Fixes #170 
Basically caused by inefficient highlight queries.


fdncred:
cool. thanks

PR #172, opened 2024-12-18
fix: parameter comment doc highlight

blindFS:
The doc comment string of a parameter is actually parsed as a child node

Forgot to check it in #171 .


fdncred:
Thanks

PR #176, opened 2024-12-28
fix: comma in `_match_pattern_record` and `_match_pattern_list`

blindFS:
This PR fixes parsing error caused by comma in `_match_pattern_record`:

image

as well as the extra item of anonymous comma in `_match_pattern_list`:

image

also renamed `item` into `entry` to align with `list_body`


fdncred:
Thanks

PR #178, opened 2024-12-29
fix: allow `val_interpolated` as key in `record_entry`

blindFS:
This PR fixes the following parsing error.

image


fdncred:
Thanks

PR #180, opened 2025-01-27
fix: overlay use/hide

blindFS:
fixes some failed cases of overlay

```nushell
overlay use foo.nu
overlay hide -k foo
```


fdncred:
Thanks

PR #181, opened 2025-02-05
feat!: new operator `has` and `not-has`, deprecate `stmt_register`

blindFS:
This is a breaking change, as it removes `register` keyword from highlights.scm


fdncred:
Thanks

PR #182, opened 2025-02-12
feat!: attributes

blindFS:
Adds support of https://github.com/nushell/nushell/pull/14906#issuecomment-2652223446 @Bahex

image

This is a general implementation for all kinds of attributes,
We can discuss whether we need more detailed ones for specific attribute like `example`

And this is a breaking change since it modifies `highlights.scm` and adds a new `folds.scm` file,
I'll update nvim-treesitter latter when we nail down the details.


blindFS:
@Bahex are we cool to land this?

Bahex:
Played around a little with it, looks great!

fdncred:
Thanks

PR #184, opened 2025-02-22
revert: internal commands with spaces

blindFS:
In case we might want to revert #161 as [discussed](https://github.com/zed-industries/extensions/pull/2068#issuecomment-2675458136)


fdncred:
100% agree about semantic tokens.

I'm happy to revert this when you're ready. Not sure if you want to wait for max to chime in on the other thread or just pull the trigger. Let me know.

fdncred:
Thanks. Looking forward to seeing (1) how we handle commands with spaces and (2) semantic tokens in the future.

Issue #185, opened 2025-02-22
Tracking issue: performance optimization


blindFS:
As mentioned in https://github.com/zed-industries/extensions/pull/2068#issuecomment-2675494003, current parser is fairly complicated, and needs to be optimized.

# Status

## State Count

```sh
> rg "#define.*STATE" src/parser.c
```

> 16:#define STATE_COUNT 7954
> 17:#define LARGE_STATE_COUNT 1749

## Top Results of States for Rule

```sh
tree-sitter generate --report-states-for-rule -
```
| rule name | count |
|----------|-------|
|val_table                 | 1107 |
|val_range                 | 817  |
|expr_binary_parenthesized | 614  |
|shebang_repeat1           | 573  |
|_immediate_decimal        | 474  |
|_val_range                | 374  |
|val_table_repeat1         | 363  |
|_val_number_decimal       | 362  |
|collection_type           | 330  |
|_multiple_types_repeat1   | 272  |
|_val_range_with_end       | 266  |
|expr_binary               | 210  |
|collection_type_repeat1   | 189  |
|val_closure               | 183  |
|unquoted                  | 174  |
|_unquoted_anonymous_prefix| 166  |
|_unquoted_with_expr       | 165  |
|ctrl_if_parenthesized     | 158  |
|_str_double_quotes        | 156  |

 # References

https://github.com/tree-sitter/tree-sitter/wiki/Tips-and-Tricks-for-a-grammar-author#reducing-state-count

fdncred:
It may be worth tracking building with WASM. Last run for me was `6min 49sec`.

blindFS:
> It may be worth tracking building with WASM. Last run for me was `6min 49sec`.

I got these error messages, do you know where's the problem?

```
▓ 󰊠 󰏫 tree-sitter build --wasm
WARNING: image platform (linux/amd64) does not match the expected platform (linux/arm64)
[...]

PR #187, opened 2025-02-23
refactor: body rules for states count reduction

blindFS:
Some low hang fruit of #185 

After:
> 16:#define STATE_COUNT 6505
> 17:#define LARGE_STATE_COUNT 1570


fdncred:
wow, that's a big jump to 6505. nice work!

PR #189, opened 2025-03-01
refactor: unquoted_pattern_repeat & flag_equals_value

blindFS:
#185 

> 16:#define STATE_COUNT 6432
> 17:#define LARGE_STATE_COUNT 1254

The case statement count in `ts_lex` reduced from 4040 to 3290, while the total state count doesn't change much.
Time to verify the statement that wasm compilation time is mainly determined by `ts_lex` case count.

Sadly, `tree-sitter build -w` keeps pulling the docker image of wrong arch on my arm64 machine.
Have to rely on your test number, @fdncred. 

One more failing case compared to previous ver:

```nushell
echo e>foo
echo err>>foo
```

But I checked `nu_scripts`, seems nobody actually writes crazy stuff like that

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1,3 +1,4 @@
+const f = 1;
 /// <reference types="tree-sitter-cli/dsl" />
 // @ts-check
 module.exports = grammar({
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks. I'll post timing in the tracking issue.

PR #192, opened 2025-03-10
fix: id in collection_type should allow hyphen

blindFS:
Fixes parsing error in this:

```nushell
let x: record<foo-bar: string> = { foo-bar: 'tree-sitter' }
```

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -287,7 +287,13 @@ module.exports = grammar({
       ),
     _collection_entry: ($) =>
       seq(
-        field("key", choice($.identifier, alias($.val_string, $.identifier))),
+        field(
+          "key",
+          choice(
+            alias($._unquoted_in_record, $.identifier),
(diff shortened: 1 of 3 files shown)


fdncred:
conflicts now, probably due to landing your other pr.

blindFS:
> conflicts now, probably due to landing your other pr.

Yeah, easy to cause conflict with parser.c in the repo.

fdncred:
Thanks

blindFS:
@fdncred I wonder why hyphen is banned in variable names?

image

to distinguish from `1 - 1`?

fdncred:
> to distinguish from 1 - 1?

ya, i believe that's it. sorry, i forgot about that. also, i forget where but there is some mechanism that automatically changes a-b into a_b. I thought it was custom commands or custom command parameters but those aren't allowed now.

PR #194, opened 2025-03-11
fix: empty closure

blindFS:
Fixes

```nushell
{||}
```

Seems to me that rules for `val_closure`/`val_record`/`block` are delicately balanced in current parser.
That's not very healthy for maintenance. 

Each time I tried to optimize those rules, I ended up breaking some test cases.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -33,11 +33,12 @@ module.exports = grammar({
     [$._expression_parenthesized, $._expr_binary_expression_parenthesized],
     [$._match_pattern_list, $.val_list],
     [$._match_pattern_record, $._value],
+    [$._match_pattern_record, $.val_record, $.val_closure],
     [$._match_pattern_record, $.val_record],
     [$._match_pattern_value, $._value],
     [$._parenthesized_body],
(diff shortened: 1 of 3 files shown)


fdncred:
> Seems to me that rules for val_closure/val_record/block are delicately balanced in current parser.
That's not very healthy for maintenance.

Any ideas how to fix this?

fdncred:
I guess github is having some problems. it's not letting me merge.
image

fdncred:
weird. it finally merged i guess. thanks.

blindFS:
> > Seems to me that rules for val_closure/val_record/block are delicately balanced in current parser.
> 
> That's not very healthy for maintenance.
> 
> 
> 
> Any ideas how to fix this?

Sadly no, I get quite confused by the behavior of the parser here.

One possible (but not ideal) direction would be to combine closure and block into one single rule.
The drawback is that it can't tell the difference then. Hopefully it won't affect simple tasks like highlighting/indenting/folding.

alpoz11 (approved):

alpoz11 (approved):

PR #202, opened 2025-06-05
fix: allow flag equals immediate subexpressions

blindFS:
Fixes the error caused by such syntax: `foo --bar=(baz)`.


blindFS:
@fdncred This should be ready to merge.

fdncred:
Thanks!

Issue #199, opened 2025-05-12
New type of bare word string interpolation


blindFS:
Should enable `echo (2)fork` if https://github.com/nushell/nushell/pull/15735 gets merged.

Seems not very straightforward to fix, raised as an issue.

fdncred:
looks like it was merged

Juhan280:
Will you please start working on this, topiary-nushell is not working because I used the new syntax in my scripts

blindFS:
> Will you please start working on this, topiary-nushell is not working because I used the new syntax in my scripts

I'd suggest to use the equivalent quoted syntax in your scripts. I feel it extremely difficult to lex such conflicts in tree-sitter grammar.

PR #204, opened 2025-06-17
fix: new syntax in ver.105

blindFS:
Adds support for the new syntaxes of nu 0.105:

1. where + variable predicate
2. oneof type constructor

cc. @Bahex, please remind me if I missed something.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -274,10 +274,16 @@ module.exports = grammar({
       ),
 
     _type_annotation: ($) =>
-      field("type", choice($.list_type, $.collection_type, $.flat_type)),
+      field(
+        "type",
+        choice($.list_type, $.collection_type, $.flat_type, $.composite_type),
+      ),
 
     _all_type: ($) =>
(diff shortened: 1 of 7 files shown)


Bahex:
Apart from these there is the bang/`!` suffix for cell-path members, but I think the current parser doesn't have a problem with it. :+1:

blindFS:
> Apart from these there is the bang/`!` suffix for cell-path members, but I think the current parser doesn't have a problem with it. :+1:

I see, will leave that to another PR together with the cellpath literal fix.

132ikl:
i think i saw an issue with the new/fixed bare word interpolation

132ikl:
i think i saw an issue with the new/fixed bare word interpolation, ill see if i can reproduce that

blindFS:
> i think i saw an issue with the new/fixed bare word interpolation, ill see if i can reproduce that

You mean #199 ?

Issue #200, opened 2025-05-28
Cell-path literals are not recognized


Bahex:
Example input:
```nushell
[$. $.a $.b]
```

Parser output:
```
(nu_script [0, 0] - [1, 0]
  (pipeline [0, 0] - [0, 12]
    (pipe_element [0, 0] - [0, 12]
      (val_list [0, 0] - [0, 12]
        (ERROR [0, 1] - [0, 3])
        (list_body [0, 4] - [0, 11]
          entry: (val_entry [0, 4] - [0, 7]
            item: (val_variable [0, 4] - [0, 7]
              (ERROR [0, 5] - [0, 6])
              name: (identifier [0, 6] - [0, 7])))
          entry: (val_entry [0, 8] - [0, 11]
            item: (val_variable [0, 8] - [0, 11]
              (ERROR [0, 9] - [0, 10])
              name: (identifier [0, 10] - [0, 11]))))))))
```

Notes:
- cell-path literals always start with `$.`, which itself is a valid cell-path (equivalent to `/` in a filesystem)
- cell-path members may be one of:
  - integer (negative integers are syntactically allowed, but don't have clear semantics)
  - bare string
  - quoted string (can be `'`, `"`, `\``)
- cell-path members may end with `?`, `!` or a combination of both. Normally, using either of these symbols more than once for a path member is an error, but that's not relevant for the tree-sitter parser.

Example demonstrating all cases:
```
$.
$.aa."b b".0.-1.cc?.dd!.ee!?.'f f'!?.`g g`
```

PR #205, opened 2025-06-18
fix: cellpath literal as new value type

blindFS:
related to #200 

A subtle nuance

* `$.`, `[].`, `{foo: bar}.` valid in nushell, invalid in tree-sitter-nu
* `{}.` invalid in both

I don't think any of those is of practical use. And enabling that (by all means that I know) requires extra conflict resolving, which will largely increase the "case statement count", thus the WASM compilation time #185 .

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -693,14 +693,19 @@ module.exports = grammar({
 
     _blosure: ($) => choice($.block, $.val_closure),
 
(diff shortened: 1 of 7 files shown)


fdncred:
ya, we definitely need to keep the wasm compile time down for zed.

I agree, those don't seem very practical.

fdncred:
thanks. are you ready to land this?

Issue #206, opened 2025-06-18
Invalid node type "composite_type"


AdrienLemaire:
Using NixOS, nushell 0.104.1 and LazyVim in neovim.

When running `:checkhealth nvim-treesitter`, I can see an error related to the nu parser

```
==============================================================================
nvim-treesitter:                                                          1 ❌

Installation ~
- ✅ OK `tree-sitter` found 0.25.3 (parser generator, only needed for :TSInstallFromGrammar)
- ✅ OK `node` found v24.1.0 (only needed for :TSInstallFromGrammar)
- ✅ OK `git` executable found.
- ✅ OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: gcc (GCC) 14.2.1 20250322
- ✅ OK Neovim was compiled with tree-sitter runtime ABI version 15 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.15.2-zen1",
  sysname = "Linux",
  version = "#1-NixOS ZEN SMP PREEMPT_DYNAMIC Tue Jan  1 00:00:00 UTC 1980"
} ~

Parser/Features         H L F I J
  - c                   ✓ ✓ ✓ ✓ ✓
  - css                 ✓ . ✓ ✓ ✓
  - csv                 ✓ . . . .
  - diff                ✓ . ✓ . ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - javascript          ✓ ✓ ✓ ✓ ✓
  - latex               ✓ . ✓ . ✓
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - norg                . . . . .
  - nu                  x . ✓ ✓ ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - regex               ✓ . . . .
  - scss                ✓ . ✓ ✓ ✓
  - svelte              ✓ ✓ ✓ ✓ ✓
  - tsv                 ✓ . . . .
  - tsx                 ✓ ✓ ✓ ✓ ✓
  - turtle              ✓ ✓ ✓ ✓ ✓
  - typescript          ✓ ✓ ✓ ✓ ✓
  - typst               ✓ . ✓ ✓ ✓
[...]

PR #207, opened 2025-06-19
docs(neovim)!: stop installing this repo as a neovim plugin

blindFS:
For #206 and similar issues.

As discussed on the discord, this repo should be editor agnostic and parser-focused.

Editor specific query files should be maintained by downstream plugins.

I'm planning on moving `/queries/nu/*.scm` to `/queries/*.scm`, which is the more common place used by other tree-sitter parser repos. After that, the query files have to be downloaded manually to take effect, leading to less compatibility issues.

### DIFF
--- a/installation/neovim.md
+++ b/installation/neovim.md
@@ -1,8 +1,8 @@
 # neovim
 
-## Lazy.nvim installation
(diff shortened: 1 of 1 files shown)


fdncred:
ok, thanks

PR #210, opened 2025-07-03
feat(nvim): add highlighting for regex strings and 'nu -c <commands>'

Bahex:


fdncred:
oops, i guess i landed too fast. another pr is fine for this new formatting, of course.

queries/nu/highlights.scm
```diff
@@ -275,3 +275,27 @@ key: (identifier) @property
...
+        (long_flag name: (_) @_lflag (#eq? @_lflag "regex"))
+    ]
+    .
+    arg: (_ (string_content) @string.regexp)
+))
+
+((_
+    opr: ["=~" "!~" "like" "not-like"]
+    rhs: (_ (string_content) @string.regexp)
+))
```
blindFS, line 301:
```suggestion
(command
  head: ((cmd_identifier) @_cmd
    (#match? @_cmd "^\\s*(find|parse|split|str)$"))
  flag: [
    (short_flag
      name: (_) @_sflag
      (#eq? @_sflag "r"))
    (long_flag
      name: (_) @_lflag
      (#eq? @_lflag "regex"))
  ]
  .
  arg: (_
    (string_content) @string.regexp))

(_
  opr: [
    "=~"
    "!~"
    "like"
    "not-like"
  ]
  rhs: (_
    (string_content) @string.regexp))
```

Same here. BTW I don't find this useful for nvim. Does this serve as a backup plan if injections.scm doesn't work?

Bahex: It allows styling the string as a whole in addition to the injected language. You can use a different bg color for example:
image

blindFS: I see, cool.

PR #211, opened 2025-07-03
improve language injections

Bahex:
- refactor: simplify queries and reformat
- feat: arbitrary language injection: 
  image
- allow additional formatting with `@string.code`
  image

### DIFF
--- a/queries/nu/injections.scm
+++ b/queries/nu/injections.scm
@@ -33,3 +33,17 @@
   arg: (_
     (string_content) @injection.content
     (#set! injection.language "nu")))
+
+; arbitrary language injection
+; ("sql"; r#'select * from table'#)
+(expr_parenthesized
+  .
+  (pipeline
+    (pipe_element
+      (_
+        (string_content) @injection.language)))
+  .
+  (pipeline
+    (pipe_element
+      (_
+        (string_content) @injection.content))) .)
--- a/queries/nu/injections.scm
+++ b/queries/nu/injections.scm
@@ -33,17 +33,3 @@
   arg: (_
     (string_content) @injection.content
     (#set! injection.language "nu")))
-
-; arbitrary language injection
-; ("sql"; r#'select * from table'#)
-(expr_parenthesized
-  .
-  (pipeline
-    (pipe_element
-      (_
-        (string_content) @injection.language)))
-  .
-  (pipeline
-    (pipe_element
-      (_
-        (string_content) @injection.content))) .)
--- a/queries/nu/injections.scm
+++ b/queries/nu/injections.scm
@@ -1,39 +1,35 @@
 ((comment) @injection.content
- (#set! injection.language "comment"))
+  (#set! injection.language "comment"))
 
-((command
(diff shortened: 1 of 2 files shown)


queries/nu/injections.scm
```diff
@@ -1,39 +1,49 @@
...
+  .
+  (pipeline
+    (pipe_element
+      (_
+        (string_content) @injection.language)))
+  .
+  (pipeline
+    (pipe_element
+      (_
+        (string_content) @injection.content))) .)
```
blindFS, line 49:
```suggestion
(expr_parenthesized
  .
  (pipeline
    .
    (pipe_element
      (val_string
        (string_content) @injection.language)) .)
  .
  (pipeline
    .
    (pipe_element
      (val_string
        (string_content) @injection.content)) .) .)
```

Probably better to add more restrictions.

PS: the last-child-anchor `.` doesn't work at all in my neovim, not sure why.

blindFS: I mean it still matches strange structures like:

image

blindFS:
The highlight of "sql" is kinda distracting to me. I'd prefer to hide the language names in low-profile comments.

But I can tell that you want to keep away from the `gsub` stuff here.

blindFS:
> Unfortunately we don't have block comments yet... I'll try dimming the language name and/or using conceal.

I mean like this:

image

blindFS:
> @blindFS Running [format-queries.lua](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/format-queries.lua) in the repo reformats all query files with a lot changes. I only formatted the portions I added to keep the diff small.
> 
> We should probably reformat all the query files with `format-queries.lua` or `topiary`

That's an interesting story because topiary has its own formatter for scheme query files. And the results are quite different from nvim-ts's.
I'm not sure which one is more suited for topiary queries, but for scm files in this repo, I think they will end up in nvim-ts someday, so probably their format to follow.

queries/nu/injections.scm
blindFS: > I mean it still matches strange structures like:
> 
> image

@Bahex can you reproduce this? I think that's a bug of treesitter itself.

Bahex: Yeah, I can reproduce it, even with an ridiculously strict query like this:
```tree-sitter-query
(expr_parenthesized .
  "("
  (pipeline .
    (pipe_element .
      (_ (string_content) @injection.language)
	  .)
	.) ";"
  .
  (pipeline .
    (pipe_element .
      (_ (string_content) @injection.content)
	  .)
	.)
  ")"
  .
)
```

Bahex:
`conceallevel=0`:
image

`conceallevel=2`:
image

queries/nu/injections.scm
blindFS: I've tested it on other parsers like bash, and it works as expected, seems like a parser specific one.
Could be a good thing if it leads to a pitfall of our current grammar.

Bahex:
Let's leave the arbitrary injection stuff for later, it will lead to too many false positives with the current parser behavior.

PR #212, opened 2025-07-05
feat: allow new type of bare word string interpolation

blindFS:
While fixing #199, this PR will increase case number of `ts_ls` from 3376 to 4111. #185 

Not sure we want to land this.

Personally I'd like nushell to forbid all subexpressions in unquoted strings, because that's the job of `$""`/`$''` string interpolation in the first place. But the ease/robustness of parsing it brings won't worth such a huge breaking change, probably we still need this fix in the future and find another way to reduce the compilation time.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1942,7 +1942,8 @@ function _unquoted_with_expr_rule(type) {
       excluded += '{}:,';
       break;
   }
-  const pattern_repeat = token(repeat(none_of(excluded)));
+  const pattern_repeat = token.immediate(repeat(none_of(excluded)));
+  const pattern_repeat1 = token.immediate(repeat1(none_of(excluded)));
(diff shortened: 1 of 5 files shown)


blindFS:
That makes sense, yet users may not follow the best practice.

blindFS:
Optimized a little bit, `ts_ls` cases 3376 -> 3608

WASM comp time: 3m40 -> 5m40.

fdncred:
ready to land?

blindFS:
Sorry I got the number wrong, 5min40 was from previous commit.

With `3608` cases, it takes more than 15min to compile, which is unacceptable IMO.
Closing this for now.

fdncred:
oof, 15min would probably break zed. darn.

PR #213, opened 2025-07-07
chore: CI workflow update

blindFS:
Include language binding tests and parsing statistics of [nushell/nu_scripts](https://github.com/nushell/nu_scripts)


blindFS:
Also added a publish workflow, currently only used for WASM compilation test.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+    tags: ["*"]
+
+permissions:
+  contents: write
+  id-token: write
+  attestations: write
+
+jobs:
+  github:
+    uses: tree-sitter/workflows/.github/workflows/release.yml@main
```
fdncred, line 14: Where exactly is this pointing? Is it this? https://github.com/tree-sitter/tree-sitter/blob/master/.github/workflows/release.yml

I don't really like the idea of pointing to a ci that we have no control over or can change without our input, if that's what this is doing.

blindFS: That's it, it's widely used across "official" parser repos.

It only gets called when a git tag is added, and we have none. I only use that to test wasm compilation on my forks.

fdncred: I still don't like it for the reasons I listed, but I don't have a problem taking your word that it's ok.

blindFS:
@fdncred It's ready to merge.

You can safely ignore the publish workflow, we won't trigger it in the foreseeable future.
As for the outsourced actions in `ci.yml`. That's what it takes if we want to use the official tools to help ease our test procedure, I think it's no different to using the [rust toolchain action](https://github.com/actions-rust-lang/setup-rust-toolchain) for example (Maybe there're security checks for registered actions? I'm not sure).

Most of the language binding related files are regenerated by `tree-sitter init -u`, the original ones probably are broken long time ago.

Also enabled `eslint-config-treesitter` to work along with `prettier`.

Issue #198, opened 2025-04-29
Assignment rhs includes all the white space after the assignment operator.


Bahex:
This affects `stmt_let`, `stmt_mut` and `assignment`.

![Image](https://github.com/user-attachments/assets/6c27474d-578d-42cf-9e7d-e24867dcb9e1)

Noticed this while adding nu support to [`Wansmer/treesj`](https://github.com/Wansmer/treesj)

The plugin in question keeps white space between nodes intact, but appears to trim the nodes themselves during formatting.

```nushell
do {
    let a = 1
}

# expected output of join operation
do { let a = 1 }

# actual output of join operation
do { let a =1 }
```

blindFS:
Related tree-sitter issues: https://github.com/tree-sitter/tree-sitter/issues/4091, https://github.com/tree-sitter/tree-sitter/issues/3966

For some reason, the bug tag was removed 4 days ago. Maybe the authors thought it was a grammar issue. Generally I have no idea of how to fix this. Sometimes simply reordering/refactoring the grammar rule works but other times it makes things worse.

PR #214, opened 2025-07-08
fix: unexpected spaces at token head

blindFS:
This PR:

1. reduces case count #185 : 3376 -> 3146, WASM comp time: 3m40 -> 2m20
2. accidentally fixes the space issue of #198 / hide-env / flags


fdncred:
interesting. thanks!

PR #216, opened 2025-07-08
fix(highlight): highlight operators in where predicates

Bahex:
I don't know really know the performance characteristics of tree-sitter queries (nor how to benchmark them), so while just using a wildcard was my first instinct, I used an explicit where_predicate node.

Would this be better or worse?
```tree-sitter-query
(_
  opr: _ @operator)
```

### DIFF
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -97,6 +97,9 @@ file_path: (val_string) @variable.parameter
 (expr_binary
   opr: _ @operator)
 
+(where_predicate
+  opr: _ @operator)
+
 (assignment [
     "="
     "+="


blindFS:
In [nvim-ts 's highlight.scm](https://github.com/nvim-treesitter/nvim-treesitter/blob/42fc28ba918343ebfd5565147a42a26580579482/queries/nu/highlights.scm#L132-L204), it enumerates over all possible leaf nodes.

I suppose there's diff in performance. I've never tested that. There's a `-t` tag of `tree-sitter query` command that might be helpful.

blindFS:
@Bahex According my tests, there's not much query perf diff between those two forms --- time of `tree-sitter query -t xxx xxx`.

But less rules does help in overall runtime --- `timeit {tree-sitter query xxx xxx}`.
Probably faster query parsing, which is not critical IMO.

Would you please merge this? It's something I forgot to add in #205 , thanks for the fix!

PR #218, opened 2025-07-16
fix: env_var with (interpolated) string as value

blindFS:
Fixes the parsing error described in https://github.com/blindFS/topiary-nushell/issues/20

Known difference to native nu-parser after the fix:`FOO=(bar) baz` 
1. error by tree-sitter-nu.
2. assign a string of `'(bar)'` (with the parentheses) to env variable `FOO` in nu

I guess that's a corner case nobody actually cares.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1302,14 +1302,16 @@ module.exports = grammar({
 
     env_var: ($) =>
       seq(
+        field('variable', alias($.cmd_identifier, $.identifier)),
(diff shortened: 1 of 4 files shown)


fdncred:
kind of a weird scenario to have `'(bar)'`. of course, nushell would want to evaluate that. in nushell you'd probably have to do `char lp` and `char rp`.

Issue #179, opened 2024-12-30
Should we try to minimize the amount of keywords?


blindFS:
Background:

there are many special rules of internal commands in tree-sitter-nu ATM, which have not much difference to general commands, such as

* overlay
* register
* do
* loop
* error make

This can cause errors when users accidentally shadow them with customized definitions, e.g.

https://github.com/nushell/nu_scripts/blob/a83a40dff05e91daf90bb42e7a23c5e70c85a759/sourced/misc/nu_defs.nu#L492

It is actually a common issue among tree-sitter parsers, official doc suggests using [keyword-extraction](https://tree-sitter.github.io/tree-sitter/creating-parsers/3-writing-the-grammar.html?highlight=keywords#keyword-extraction) to handle it.
But it's not feasible in tree-sitter-nu, at least for now, since we already have `identifier` as `word` and we can't ignore the subtle differences between `cmd_identifier` and `identifier`.

So, what I'm suggesting here is removing those trivial rules and reduce them all into the `command` category.
As for the highlighting, we can still make them highlighted as keywords if we want to.

fdncred:
My overall goal would be for someone to be able to use tree-sitter-nu as a nushell parser and recreate nushell if they wanted to. I'd like it to be at least that accurate.

So, I'm fine with reducing into the command category as long as it meets the first overall goal.

My second overall goal is to have excellent syntax highlighting so, i wouldn't want any change to affect that.

blindFS:
In terms of semantic accuracy/completeness, I think it's better to keep as it is.
Maybe because I never dig deep into the nushell's parser, the boundary between internal commands and the language essentials is a little bit obscure to me.

`scope commands | where category == core` generates even more stuff. Just wondering how to tell the difference.

fdncred:
I wouldn't worry about categories too much, they can change pretty easily.

- `custom` are `custom commands` like `def blah [] {}`
- `keyword` are usually created in the nushell parser aka parser keywords
- `built-in` are every other commands built into the nu binary
- `plugin` are separate binaries starting with nu_plugin_ that are currently registered
- `external` are custom completions 
- `alias` are aliases

```nushell
❯ scope commands | group-by type
╭──────────┬──────────────────╮
│ custom   │ [table 137 rows] │
│ keyword  │ [table 34 rows]  │
│ built-in │ [table 396 rows] │
│ plugin   │ [table 163 rows] │
│ external │ [table 110 rows] │
[...]

PR #220, opened 2025-07-18
refactor!: cleanup rules that can be assimilated as commands

blindFS:
The idea of #179
It helps with the WASM compilation time #185 : case count 3146 -> 2592, time 2m20 -> 1m35

IMHO the target of tree-sitter-nu is only for code text manipulation, it's never gonna be as complete/semantically accurate as the native parser. And those rules are unnecessary in that regards.


fdncred:
Big savings. I was hoping one day to see if the tree-sitter-nu parser would be faster than the nushell parser. However, if there's no hope for ever using it for anything other than and in-editor parser, then it's probably good to optimize it for that.

blindFS:
@fdncred yeah, I understand the intention. However according to what I learned from the tree-sitter community, it's generally not recommended to use a tree-sitter parser in a compiler/interpreter.

Besides, performance-wise, I believe the new-nu-parser(current nu-parser can be slow) should be by no means slower than TS. The only thing missing is the incremental nature.

queries/nu/textobjects.scm
```diff
@@ -50,15 +60,19 @@
...
 (ctrl_if) @statement.outer
+
 (ctrl_try) @statement.outer
+
 (ctrl_match) @statement.outer
+
 (ctrl_while) @statement.outer
+
 (ctrl_loop) @statement.outer
-("break" "continue") @statement.outer
```
blindFS, line 62: @Bahex I've merged `ctrl_do`/`ctrl_return`/"break"/"continue" to our `command` rule.
I'm not sure whether some features will be missing if we remove these queries, do you have any idea?

Bahex: It should be fine as far as I can tell, and even if we want to add queries for these in the future, we can just write the them for `command` nodes with a specific head.

blindFS: @Bahex Would you like to review/test it since there are some breaking changes?

blindFS:
> [`nvim-treesitter-context`](https://github.com/nvim-treesitter/nvim-treesitter-context/blob/59f318a65d42a5c4037796886a8874fd57f774fc/queries/nu/context.scm#L35) has a query that would need to be updated, but other than that I haven't found anything that breaks.

Thanks for the reminder, I haven't updated the parser-rev/query-files of nvim-treesitter since cli 0.25+ was applied in this repo.
Their main branch breaks some default configurations of the nvim distro that I'm using. I guess most users are still stuck in their frozen master branch.

blindFS:
> Should we go ahead and land this @blindFS?

I think so. Given there're already big gaps between the main branch and those locked by editors/plugins, it won't break things further.

fdncred:
Thanks

PR #225, opened 2025-09-19
fix(grammar): handle cursed env flag positions

mkatychev:
Handle function definitions where flags appear after the command name:
https://github.com/nushell/nushell/blob/9ebb02d23691a373298ac019d5081df508513733/crates/nu-command/tests/commands/def.rs#L228-L236

* updated tree-sitter-cli to 0.25.9
* added `ts regen` command to `ts.nu`
* added `linguist-generated` attributes to generated code


bindings/rust/example/main.rs
```diff
@@ -1,106 +0,0 @@
-extern crate tree_sitter_nu;
```
fdncred, line 1: Please undelete this file

test/corpus/decl/def.nu
```diff
@@ -48,6 +48,8 @@ def --env foo [] {}
 def --env --wrapped bar [...args] {}
 def --wrapped foo [...args] {}
 def --wrapped --env bar [...args] {}
+def bar --wrapped --env  [...args] {}
```
blindFS, line 51: Interesting, this case also fails the github highlighter of nushell. LoL

ts.nu
```diff
@@ -86,3 +84,23 @@ export def "ts parse" [    # -> string
 export def "ts config" [] {
     tree-sitter init-config
 }
+
+# Regenerates tree-sitter bindings and generated manifests
+export def "ts regen" [] {
```
blindFS, line 89: This is handy, thanks! Does it cause any trouble if we don't remove the existing ones BTW? I thought (probably wrong) `--update` means incremental changes.

blindFS: > I think one problem is that there can be "stuck" patterns that newer versions of `tree-sitter-cli` leave alone 

True, a whitelist of files we want to keep?

BTW, tree-sitter-cli 0.26.0 just got released, would you like to run regen with that?

mkatychev: > BTW, tree-sitter-cli 0.26.0 just got released, would you like to run regen with that?

Looks like the NPM packages aren't up to date yet (so node tests might break in CI):
https://www.npmjs.com/package/tree-sitter

we can wait for those or do that as a followup.

mkatychev: this was the output of running `tree-sitter test -r`:
image


The issue was fixed after running `cargo install tree-sitter-cli` _without_ making any changes to `grammar.js`.

blindFS (approved):

fdncred:
Thanks

PR #226, opened 2025-10-06
feat!: add support of new custom constant completer syntax

blindFS:
https://github.com/nushell/nushell/pull/16789

A breaking change since `param_cmd` is now renamed as `param_completer`.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -158,7 +158,7 @@ module.exports = grammar({
         optional(modifier().visibility),
         keyword().def,
         repeat($.long_flag),
-        field('name', $._command_name),
+        $._command_name,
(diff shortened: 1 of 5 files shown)


blindFS:
CC @Bahex

Bahex (approved):
@blindFS wow thanks for the quick update (after I added yet more syntax :pensive:)

I got confused about `val_record` in `param_completer` at first, then I saw the test case.
```nushell
def test [
  x: int@[1 2],
  y: string@{f: [a b]}.f,
  --zero(-z): list<string@$var.cell?>
= [a b],
] {}
```

Looks good to me!

PR #247, opened 2026-03-20
Expose the Highlights query at crate level

1gmar:
When tree-sitter-nu crate gets published, this will allow it to be easily integrated in other tools that use treesitter highlighting, such as Difftastic.

Issue #245


fdncred:
Thanks. BTW - I'm fine with publishing any time.

PR #248, opened 2026-04-22
add optional sigil `%` to start of commands

Tyarel8:
Add the new sigil `%` for builtin commands from https://github.com/nushell/nushell/pull/17956

### DIFF
--- a/test/corpus/expr/identifier.nu
+++ b/test/corpus/expr/identifier.nu
@@ -252,3 +252,28 @@ def 🤖👀 [] {}
     (pipe_element
       (command
         (cmd_identifier)))))
+
+====
+cmd-id-005-builtin
+====
+
+%echo
+%str join
+%each
+
+-----
+
+(nu_script
+  (pipeline
+    (pipe_element
+      (command
+        (cmd_identifier))))
+  (pipeline
+    (pipe_element
+      (command
+        (cmd_identifier)
+        (val_string))))
+  (pipeline
+    (pipe_element
+      (command
+        (cmd_identifier)))))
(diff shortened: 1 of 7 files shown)


blindFS:
Thanks for the update, would you please also include some new test cases?

https://github.com/nushell/tree-sitter-nu/blob/main/test/corpus/pipe/commands.nu

test/corpus/expr/identifier.nu
```diff
@@ -252,3 +252,28 @@ def 🤖👀 [] {}
     (pipe_element
       (command
         (cmd_identifier)))))
+
+====
+cmd-id-005-builtin
+====
+
+%echo
+%str join
```
blindFS: This doesn't sound right to me...

Maybe we need to tune the nushell syntax a little bit, `%"str join"` makes more sense to me. @fdncred ideas?

fdncred: I think the way it's listed above is the right syntax. You don't call sub commands with quotes so I'm not sure why you'd need to quote it if you're using a sigil.
```nushell
❯ [a b c] | str join
abc
❯ [a b c] | %str join
abc
```

blindFS: Yeah, good point, we can't tell whether the `join` here is part of the command or a command argument with or without the sigil. We chose the lsp semantic tokens feature as a workaround.

@Tyarel8 I'd suggest we remove this single line from the test, as the result is not aligned with nu-parser. That's something we can't fix at the moment.

Tyarel8: Alright

fdncred:
Are we ready to land this now?

blindFS:
> Are we ready to land this now?

Sure.

blindFS (approved):

fdncred:
Thanks

PR #252, opened 2026-07-01
fix for comment in interpolated string

### LINKED ISSUE #251: Parse error of multiline string interpolation with `#`
https://github.com/nushell/nu_scripts/blob/788e144a109c5f47e3a24c75448b775c7f088368/make_release/twin-tweaks.nu

Somehow it gets treated as a comment

fdncred:
This PR closes #251 by updating the `comment` regex to:
to
```js
comment: _ => seq('#', /[^)\]}\n]*/),
```
meaning comments start with `#` and stop at `)`, `]`, and `}`. The trade-off is that `)`, `]`, and `}` can no longer appear in comments on the same line as `#`. 

It's probably an acceptable trade-off.

The problem was strings like `$"foo (## some comment) bar"` weren't being interpreted correctly and the tree-sitter parser was basically breaking and didn't know what to do.

- I also made some minor changes the the package.json so `tree-sitter test` would work better for me.
- I updated the .gitignore to ignore .DS_Store on macs
- I also added a test for string interpolated evaluated comments evaluating to nothing. LOL
- I had to set `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` to true in the CI to get this to work since there's no new action that uses Node 24 yet.


fdncred:
ugh, apparently this fix breaks other things. i'll look a bit more before giving up. :)

blindFS:
Should be much easier if there's [this](https://github.com/tree-sitter/tree-sitter/issues/931) feature of ts

PR #258, opened 2026-08-25
fix(scanner): avoid out-of-bounds read of valid_symbols during error recovery

### LINKED ISSUE #257: Out-of-bounds read (global-buffer-overflow) in the external scanner
## Summary

- Repository: https://github.com/nushell/tree-sitter-nu
- Commit tested: `d694570aa26b53d0d642460a0430e8aa07dcbea0`
- Component: external scanner (`src/scanner.c`)
- Class: **global-buffer-overflow**
- Impact: the external scanner indexes past a global buffer (AddressSanitizer) — potential memory corruption. Any tool that parses untrusted input with this grammar (editors, language servers, syntax-highlighting services) is exposed.

## Reproduce

Because this is an unpatched memory-corruption bug, I'm not publishing a working
reproducer here. I have a self-contained one — a build script plus the crashing input,
pinned to the commit above — and I'm glad to share it privately with a maintainer (or
attach it to the fix PR once a fix is in progress). Please reach out and I'll send it over.

## Root cause

The crash is an out-of-range read of `valid_symbols[ERROR_SENTINEL]` (around line 116 of `src/scanner.c`): `ERROR_SENTINEL` does not correspond to a real external token slot — the scanner declares only 3 tokens in `externals`, so `valid_symbols` is sized for 3 — so indexing it with `ERROR_SENTINEL` reads past the end of the array.

## Fix

I have a fix for this and will open a pull request shortly.

## Provenance

Found with an automated libFuzzer + AddressSanitizer fuzzing pipeline, as part of a
research project on hardening tree-sitter grammars. Happy to coordinate on disclosure timing.

xcthulhu:
Fixes https://github.com/nushell/tree-sitter-nu/issues/257

## What

Fixes a **global-buffer-overflow** in the external scanner (`src/scanner.c`).

## Root cause

The scanner reads `valid_symbols[ERROR_SENTINEL]` (around line 116), but `ERROR_SENTINEL` is not a declared external token — only 3 tokens are declared in `externals`, so `valid_symbols` is sized for 3 and the index reads past the end of the array.

## Fix

Removes the invalid `valid_symbols[ERROR_SENTINEL]` access and instead detects error recovery by checking only the real, in-range token slots (`RAW_STRING_BEGIN` / `RAW_STRING_CONTENT` / `RAW_STRING_END`).

The patch was generated with LLM assistance and **auto-verified**: after applying it and
rebuilding at the tested commit, the crashing input no longer triggers the bug when replayed
under the same libFuzzer + AddressSanitizer time / memory budget. Please review it on those
terms — it may include incidental reformatting.

## Provenance

Found with an automated libFuzzer + AddressSanitizer fuzzing pipeline, as part of a research
project on hardening tree-sitter grammars. A self-contained reproducer is available on
request (see the linked issue).

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -113,7 +113,13 @@ bool tree_sitter_nu_external_scanner_scan(
     TSLexer *lexer,
     const bool *valid_symbols
 ) {
-    if (valid_symbols[ERROR_SENTINEL]) {
+    // During error recovery tree-sitter marks every real external token as
+    // valid; ERROR_SENTINEL is not a declared external token, so indexing
+    // valid_symbols[ERROR_SENTINEL] reads past the end of the array. Detect
+    // error recovery by checking only the valid, in-range token slots.
+    if (valid_symbols[RAW_STRING_BEGIN] &&
+        valid_symbols[RAW_STRING_CONTENT] &&
+        valid_symbols[RAW_STRING_END]) {
         return false;
     }
 


fdncred:
Is it possible to have tests that support these changes?

xcthulhu:
> Is it possible to have tests that support these changes?

The failure is only reproducible with AddressSanitizer; I could incorporate it into your git workflow if you want.

fdncred:
Never heard of it. Do you have a link to it where I can read up on it? I doubt we'd want to add it to our CI.

